PR #16, opened 2016-02-24
Change bootstrap.py to respect `CXX` variable

Nava2:
The current bootstrap script assumes `g++` when compiling, instead this reads the `CXX` environment variable as most systems do.


nabijaczleweli:
:+1:

PR #18, opened 2016-02-25
Add usage of LUA_VERSION variable when using --ci

Nava2:
* Adjust finding Lua for Linux and OSX
* Readd `install.deps.sh`

There are failing tests, but they are due to code problems. 

Lots of environments and tests now. :+1: 

https://travis-ci.org/Nava2/sol2/builds/111812312

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,25 @@
-obj/*
-bin/*
-demacro.txt
-Shinobi2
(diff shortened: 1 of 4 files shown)


ThePhD:
Ooh, I didn't see that! My bad. Merging those too...

Issue #22, opened 2016-03-01
Python vs. CMake for Build tool


Nava2:
I'm trying to fix the python bootstrap.py, but I feel like I'm reinventing the wheel trying to have it find Lua vs. LuaJIT. 

Is there a reason for using a custom python script _over_ something like CMake? CMake will _happily_ [generate Ninja build files](https://cmake.org/cmake/help/v3.0/generator/Ninja.html).

nabijaczleweli:
Yes, there is. CMake is utter shit.

Nava2:
That's neither an answer, nor helpful.

nabijaczleweli:
It is *the* answer The Lounge (and, by membership, ThePhD) will give you

ThePhD:
Erm... Lounge affiliation or not, as I explained in #23, using CMake is not practical because this library doesn't have any build steps.

ThePhD:
Out of curiosity, what're you trying to "build" with sol? Is it just the automation of finding the right lua version to get bundled with it? IMHO, I would like that to be the user's choice, and not a feature (and eventual problem) we have to graft onto the library.

Nava2:
My research (for my masters.. not as the PhD level :wink:) has Lua as a primary component and needs to be tested and run cross-platform. So, finding Lua is always a pain. This is something CMake does _really_ well.

Nava2:
Oh, and I just re-read what you wrote about Lua versions, you're correct. The Lua Version is only required when building the tests or examples. Otherwise, it never asks nor requires it. That being said, I would figure Intellisense and other tools when _developing_ sol2 would have some problems if you didn't.

ThePhD:
After thinking about this a bit more, I think I'm going to have to stick with the python. The primary reason is that it's simple and will continue to be for automating our test suite. I also want to be careful because, when we hit the 2.0.0 tag (probably in the next 2 weeks), what the release story will be is literally is:

"We ran single.py and you just need to download the amalgamated `sol.hpp` to get going. Put it anywhere you like. Enjoy."

CMake doesn't really help with that story, so in the end it'd just be something that we use to run our tests... so then we have to ask ourselves "is this something we want for running the tests / examples?"

And at the moment... nah, not really. Though I can see how easy it is to get going, I am unfortunately not very good at CMake and am focusing on writing the docs / preparing for v2.0.0 release. We'll keep the python zaniness for now, and just request that users who want to do things like find the right Lua version should consult their specific platform / build system for it.

[...]

PR #23, opened 2016-03-02
Proposed change to use CMake Build

Nava2:
# Justifications

* CMake is a colloquial standard within the C++ community
* While the syntax sucks, there is a lot of support and modules available
* Allows for "automated" finding of Lua to compile against
* Removes custom logic used in `bootstrap.py` in favour of cmake's built-ins
* Catch no longer needs to be stored within a submodule

# Building

```bash
mkdir build && cd build
cmake .. -DSOL2_LUA_VERSION=lua52 -G Ninja
ninja check    # Builds/runs catch tests
ninja examples # Builds examples
ninja install  # Installs the library into the directory CMAKE_INSTALL_PREFIX
```

If `-G Ninja` is not specified, it'll use `make` by default. 

On windows, for Visual studio 2015 project files (done in the project directory):

```
cmake ./ -DSOL2_LUA_VERSION=lua52 -G "Visual Studio 14 2015 Win64"
```

## Arguments

* `SOL2_LUA_VERSION`: This specifies what Lua version to look for, acceptable patterns: `lua`, `lua51`, `lua52`, `lua53`, `luajit`, `luajit51`, `luajit52`, `luajit53`

## Options

* `SOL2_BUILD_TESTS`: set to `OFF` to not build tests
* `SOL2_BUILD_EXAMPLES`: set to `OFF` to not configure examples

# Misc

* 2a2aeac cleaned up the `.travis.yml` file a bit
* ceb960e Removes the original python build


Nava2:
While this is a big change, I am already using this in my work and it works better for me -- that being said, I know how **awful** CMake's syntax can be (they should have [just used lua](https://cmake.org/Wiki/CMake:Experiments_With_Lua)). But, the benefits of having a mature and stable project backing a build is enormous.

ThePhD:
`Sol` is a header only library: the python bootstrap is essentially exclusively for running travis-ci and (for the moment) generating a single header with single.py. "Install" and "Uninstall" are really convenience bits that, while nice, should probably be gutted (it provides an opinion about where to put things when really the only necessary framework is "get header files, add them to include list").

While I am happy to investigate ways to remove Catch as a submoduled dependency (as this gets in the way of just doing a git pull and getting running without the extra muck), CMake is overkill for our purposes. `Sol` specifically does not care about the Lua framework you are using, is not really supposed to be "installed", and has no real build steps.

Nava2:
Few things: 

1. The build is still required for building the examples and the tests -- particularly the integration tests with the versions of lua
2. Embedding the project (as it is supposed to be) is the idea, but doing that by hand without a submodule is annoying. Here is some sample code I use to embed this branch in my research work: 
  ```cmake
# Download and "install" sol2
set(SOL2_BRANCH cmake-build) # usually develop

ExternalProject_add(
    sol2
    PREFIX ${VENDOR_PREFIX}
            
    GIT_REPOSITORY "https://github.com/Nava2/sol2.git"
    GIT_TAG ${SOL2_BRANCH}
    
    CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${CMAKE_BINARY_DIR}
               -DSOL2_LUA_VERSION=${LUA_VERSION}
               -DSOL2_BUILD_EXAMPLES=off
               -DSOL2_BUILD_TESTS=off
    TEST_COMMAND "")
  ```
  While it may look a bit odd, this simple code handles downloading, and installing the library into a consumable directory for a project (and keeps it updated as branches move). Previously, I had a bash script that did the git commands, checkout, copy the folders over. Needless to say, brittle and I had to be responsible. 
3. CMake allows users to embed the application into existing cmake projects and pack it using their tooling (e.g. creating installers on Windows, deb/rpm on Linux, or dmg on OSX)

For what it's worth.. I hate CMake's language decisions, but I love how mature and simple it is (most times).  

All-in, this is very much ultimately your project, I'm just trying to sway a vote. :+1:

Nava2:
(by the way, the build fails for something due to luajit on osx, nothing I did..)

Nava2:
Oh, and IDE support for CMake is [excellent](https://cmake.org/Wiki/CMake_Editors_Support).

ThePhD:
Closed as per #22 discussion: but we will open it again if it comes imperative for us to utilize a better build system to make this work.

Thank you, still, for the work.

alkino:
I only see this now. I work on a new version of that because yesterday @ThePhD say on gitter:
> I have to learn CMake and make the appveyor and travis builds run using CMake
> Because I need to generate a project for Visual Studio 2015, 2017 to run on AppVeyor
> Then I need to generate a MiNGW project on that same Windows instance
> Then I need to have things work for all Linux systems,
> and all Mac sys- see where this is going?

PR #34, opened 2016-03-14
Removed stray <feff> byte order mark

patrikhuber:
This fixes compiling with clang on msys2. I believe you saved these files as UTF16 once or something and these byte order marks were added by an editor.
The PR leads to the line "﻿// The MIT License (MIT)" to no longer be present in the single-header sol.hpp at 3 occasions, but I think they were there by mistake anyway and should've been stripped by the python script, but didn't, because of the BOM. So this actually fixes that as well.

### DIFF
--- a/sol/function.hpp
+++ b/sol/function.hpp
@@ -1,4 +1,4 @@
-﻿// The MIT License (MIT)
+// The MIT License (MIT)
 
 // Copyright (c) 2013-2016 Rapptz, ThePhD and contributors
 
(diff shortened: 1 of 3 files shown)


patrikhuber:
The change is actually not visible in the GitHub diff, but if you inspect the old files with for example `vim -b sol/function.hpp`, you'll see the `<feff>`.

Nava2:
Thanks for fixing this, I found the single header didn't work. I assume this was why. 

[BOM](https://en.wikipedia.org/wiki/BOM) in case you're curious..

ThePhD:
Curses, Visual Studio has smacked me again.

Thanks for the fix: I'll re-generate the sources and re-tag version 2.0!

patrikhuber:
Cool! :)

PR #46, opened 2016-03-29
Added table::get_with_default method

zwparchman:
There is now a sol::table::get_with_default method that takes 2
arguments, one being the key to search on, the other being the default.
If the key does not have a value in the table the default value is
returned. I expect this to be particularly useful when using lua to optionally
override default values in c++ code.


ThePhD:
Looks nice. I'm going to change the name to `get_or`. I also need to consider how this interacts with multi-get, and with usertypes that always return references to what's in the lua VM:

````
lua.get_or<int, int>( "bark", ?? );
lua.get_or<some_user_class>( "bark", ??? );
````

PR #109, opened 2016-05-27
Use Boost Optional if USE_BOOST is defined

gsauthof:
Usecase: Your project (where you plan to to use sol2 for interfacing with lua) already uses Boost as dependency, your compiler doesn't come with <optional> yet and you want to minimize further dependencies.

That means that with that change and defining `USE_BOOST` one doesn't need to pull the solarized optional repository anymore.

### DIFF
--- a/sol/optional.hpp
+++ b/sol/optional.hpp
@@ -24,6 +24,8 @@
 
 #if __cplusplus > 201402L
 #include <optional>
+#elif defined(SOL_USE_BOOST)
+#include <boost/optional.hpp>
 #else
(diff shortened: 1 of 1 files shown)


ThePhD:
This looks nice. I'll have to change the define to be `SOL_USE_BOOST`. Apologies if it takes me a while to merge this: I'm busy refactoring the core usertype implementation for LODES-A-SPEED.

gsauthof:
Ok, I've just rebased the pull request branch such that the guard uses `SOL_USE_BOOST`.

PR #121, opened 2016-06-20
Use nullptr when converting to std::function if value is nil

chingjun:
Hi, I've just started to use sol in my own project. I ran into this problem when converting a lua function to std::function, there was no way to tell if the function is nil in C++ code without using sol::function, and the program would crash when the function is called.

### DIFF
--- a/sol/function.hpp
+++ b/sol/function.hpp
@@ -105,6 +105,9 @@ namespace sol {
 
 			template<typename... Args, typename... Ret>
 			static std::function<Signature> get_std_func(types<Ret...>, types<Args...>, lua_State* L, int index = -1) {
+				if (lua_isnil(L, index) || lua_isnone(L, index)) {
+					return nullptr;
+				}
 				sol::function f(L, index);
 				auto fx = [f, L, index](Args&&... args) -> meta::return_type_t<Ret...> {
 					return f.call<Ret...>(std::forward<Args>(args)...);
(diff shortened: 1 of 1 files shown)


Nava2:
When you get the object from sol, couldn't you do a `valid()` check before conversion? That is, if you're worried. The overhead you're adding here is not trivial.

chingjun:
@Nava2  Thanks for your reply! My situation is similar to the code below, where a lua script is calling a C++ function that accepts a callback directly but also accepts a nullptr callback, in which case it will not call the callback.

```C++
#include "sol.hpp"

void test(std::function<void()> cb) {
    printf("cb is nullptr? %d\n", cb == nullptr);
    if (cb != nullptr) {
        cb();
    }
}

int main() {
    sol::state L;
    L.set("test", test);
    L.script("test(nil)");
    return 0;
}
```

When using sol, it is impossible to check for the validity of `cb` in function `test`, and the program will crash. Is there any other way to test for the validity of the callback?

ThePhD:
It would probably make sense to merge this. My initial reaction would be to just take a `sol::function` instead and do a `valid` check as @Nava2 said, but this way allows for a bit more flexibility. And flexibility is generally nice.

ThePhD:
As a side note, I do loathe these semantics for C++: a `null` std::function is... kind've wonky, to be perfectly honest. I'm not sure why the committee included it. Probably because you can have null function pointers, but it would have made for sense for someone to have to just check it beforehand so you wouldn't have to check at the callsite all the time.

Oh well. vOv

Nava2:
Side note, should not have been gated via `SOL_CHECK_ARGUMENTS`? 

I didn't fully understand your question, @chingjun. Your use case makes sense. :+1:

ThePhD:
What do you mean by "gated"? If you mean "threw an exception", yes it actually did. I already changed `stack::check` to allow for `nil` / `none` as arguments, and those will just become `nullptr`'s for the `std::function`.

PR #144, opened 2016-07-16
fixed links to docs

reworks:


ThePhD:
Thanks for that. I wonder when that'd become necessary!

PR #150, opened 2016-07-25
fix std::nullptr_t getter and pusher

THE-FYP:
according to http://sol2.readthedocs.io/en/latest/tutorial/ownership.html
it should be that way


ThePhD:
.... Huh. I really thought I had this somewhere already. Maybe I deleted it when moving things around; so sorry about that!

Issue #157, opened 2016-08-02
Automatically bind Base Class Metatables to Derived Classes


ThePhD:
As it stands, derived classes need to re-specify base members already bound. It would make sense if there was some degree of retroactive backwards-compatible lookup for base functions so that a person did not need to register them explicitly.

Performance might be a concern at first, but just having the feature might be more worth it.

ThePhD:
https://github.com/OrfeasZ/sol2/tree/member-inheritance

Remember to look through this more thoroughly.

ThePhD:
@OrfeasZ So, I took a stab at implementing my member inheritance and added a benchmark (just for Sol2 right now, will do the other frameworks later). For my version, the cost looks like this:

![Not too bad, I guess?](http://i.imgur.com/8pmYEfk.png)

I'm using [lua bench](https://github.com/ThePhD/lua-bench) to do my benchmarking. In a moment, I'll pull your version of sol2 and run it to see what the performance looks like.

As a side note, note that my implementation doesn't require any state held C++-side, though it ostensibly costs us in lookup time (e.g., see the execution time of just calling a member function we know is there):

![Pretty decent](http://i.imgur.com/blsTuvS.png)

OrfeasZ:
Keep in mind that my branch still has the copy constructor issue I reported a few days ago. Not sure if that changes anything.

ThePhD:
Problem with my approach, as well:

Calls to a base function cast to `Base*` to do the call from the `void*` stored as a Lua userdata. This means that vtables and other important information, if that `void*` was originally a `Derived*`, get misaligned. This results in funky accesses and bad crashes at horrible times.

Not sure how to fix.

ThePhD:
And I've fixed it by keeping the casting logic in the `getter` for usertype. So everything seems to work out now.

ThePhD:
Haaaahahaha it's terrible *sobbing noises*

![SOB.](http://i.imgur.com/B9ODA3e.png)

ThePhD:
We'll release it but this issue stays open 'till we figure out a better way to do it.

OrfeasZ:
Can you share the benchmarking code you used so I can take a look when I have the chance?

ThePhD:
https://github.com/ThePhD/lua-bench

There's a sol submodule -- after cloning (and recursively init / update submodules), just point the sol submodule at your fork's url and the commit.

OrfeasZ:
[...]

PR #165, opened 2016-08-06
Support for Lua 5.3 UTF8 Library

OrfeasZ:
This patch adds support for the new Lua 5.3 library `utf8`.
One can now enable it like any other library using `sol::state_view::open_libraries(...)`:

```cpp
state.open_libraries(sol::lib::utf8);
```

### DIFF
--- a/sol/state_view.hpp
+++ b/sol/state_view.hpp
@@ -41,6 +41,7 @@ namespace sol {
 		io,
 		ffi,
 		jit,
+		utf8,
 		count
 	};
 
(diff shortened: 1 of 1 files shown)


ThePhD:
:ok_hand: Thanks, keeping sol2 up to date with small things like this is really nice of you!

OrfeasZ:
No problem! Last commit had a small typo which I fixed, so upcoming tests should all pass. I could also start looking into #157 if you want.

ThePhD:
#157 is going to be no easy feat, and there's a performance metric I need to stay in line with or improve. By all means, give it a try, but I'll be a lot more picky about how it all works out!

OrfeasZ:
Of course, this is your project anyway, so you always get the final say.
I'll peek into it a bit, see what I can come up with (since I want to get a bit more familiar with the inner workings of sol anyway).

Nava2:
Should the enum and the function be gated behind the 53 macro? Isn't it an error to require it without lua 53?

Nava2:
Rather than silently passing, rather.

PR #175, opened 2016-08-12
Fix 'expected .. received ..' message for function arguments

looki:
For some reason, whenever type::function was expected, the argument order given to the checker is the other way around. This produces messages such as:
> expected number, received function

when executing process_callback(100), whereas process_callback takes a sol::function as argument.

 I'm not sure whether this is the case for any other types, but I didn't find any cases when I had a quick look.

PS I'm absolutely loving this library so far!

### DIFF
--- a/single/sol/sol.hpp
+++ b/single/sol/sol.hpp
@@ -4607,7 +4607,7 @@ namespace sol {
 					return true;
 				}
 				if (t != type::userdata && t != type::table) {
-					handler(L, index, t, type::function);
+					handler(L, index, type::function, t);
 					return false;
 				}
 				// Do advanced check for call-style userdata?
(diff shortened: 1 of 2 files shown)


ThePhD:
Derp. I'm an idiot. Thanks for checking that: copy-paste programming flips me the bird once again..

looki:
Ugh, and I'm an idiot, too. I just realized I edited the auto-generated single header file! Do you want me to fix the appropriate header?

looki:
I'm an idiot, too. I was totally oblivious to the fact that I edited the auto-generated header that will be overwritten anyway.

ThePhD:
SNRK. I'll shift the fixes over to the main header.

There's another "I'm an idiot" comment to be made here. Something about not even realizing it was in the auto-generated header and being the repo owner... :B

looki:
It's actually in both, so that's not even necessary :laughing:

PR #183, opened 2016-08-17
Fix #180

### LINKED ISSUE #180: sol must compile with c++17 without single/sol/sol.hpp ?
Since I update the code, found compile fail, new code use such as std::in_place_tag, but my xcode config is compile with c++14.

Nava2:
This fixes problems building on OSX without the `single.hpp`. 

Additionally, I made `single.py` always use `/` instead of `\`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ x64/
 *.sln
 *.gitattributes
 
+# VSCode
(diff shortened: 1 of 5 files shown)


Nava2:
Rebased with `develop` branch.

Issue #184, opened 2016-08-18
compile with luajit 2.0.4 fail


aiplay:
I compile a cpp file with luajit instead of lua, but failed.
My compile command is : **g++ test.cpp -Isolpath -std=c++14 -lluajit-5.1**

Here is the error stack:
```
Undefined symbols for architecture x86_64:
  "_luaL_setfuncs", referenced from:
      sol::stack::pusher<sol::usertype_metatable<B, std::__1::integer_sequence<unsigned long, 0ul, 1ul, 2ul, 3ul, 4ul>, char const (&) [5], A* (B::*)(), char const (&) [4], sol::overload_set<void (B::*)(A const*), void (B::*)(char const*)>, char const (&) [5], void (*)(), char const (&) [4], sol::constructor_list<sol::types<> >&, char const (&) [5], sol::destructor_wrapper<void> const&>, void>::push(lua_State*, sol::usertype_metatable<B, std::__1::integer_sequence<unsigned long, 0ul, 1ul, 2ul, 3ul, 4ul>, char const (&) [5], A* (B::*)(), char const (&) [4], sol::overload_set<void (B::*)(A const*), void (B::*)(char const*)>, char const (&) [5], void (*)(), char const (&) [4], sol::constructor_list<sol::types<> >&, char const (&) [5], sol::destructor_wrapper<void> const&>&&) in test-690367.o
  "_lua_absindex", referenced from:
      sol::stack_reference::stack_reference(lua_State*, int) in test-690367.o
  "_lua_compare", referenced from:
      sol::stack::get_call_syntax(lua_State*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) in test-690367.o
  "_lua_getglobal", referenced from:
      void sol::stack::field_getter<char const*, true, false, void>::get<char const*&>(lua_State*, char const*&&&, int) in test-690367.o
  "_lua_pcallk", referenced from:
      sol::stack::script(lua_State*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test-690367.o
  "_lua_setglobal", referenced from:
      void sol::stack::field_setter<char const*, true, false, void>::set<char const*&, sol::stack_reference&>(lua_State*, char const*&&&, sol::stack_reference&&&, int) in test-690367.o
[...]

PR #186, opened 2016-08-21
Fix OSX + Luajit

wtfbbqhax:
First commit gives you a build.ninja that will actually link with luajit (it also means you specify luajit not luajit52 which makes more sense IMO).

Second commit makes the tests pass. Relevant to #184 if you check the luajit [installation guide](http://luajit.org/install.html) you'll find this:

```
If you're building a 64 bit application on OSX which links directly or indirectly against LuaJIT, 
you need to link your main executable with these flags:

     -pagezero_size 10000 -image_base 100000000
``` 
There really isn't an explanation for why this magic is necessary- but without it, your application _will_ crash. If you use `-fsanitize=address` you're app becomes explosive.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -110,6 +110,7 @@ elif 'darwin' in sys.platform:
         # luajit
         lua_incl = 'luajit-2.0'
         lua_lib = 'luajit'
+        ldflags.extend(['-pagezero_size 10000', '-image_base 100000000'])
     else:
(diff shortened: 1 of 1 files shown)


wtfbbqhax:
I expected that Travis was already setup to run luajit/osx but this is not the case.

ThePhD:
There's a lot of configs we don't run on travis because setting it up is a bit of a chore.

PR #188, opened 2016-08-22
Add test case for require_file with a usertype

Nava2:
I was having problems with usertypes and require_file, so I figured I should work in a less complicated testbed than mine, and worked here. There as not a test case, so I wrote this one.

Additionally, adjusted bootstrap.py to allow for `--lua-version lua53` syntax since most older linux systems use lua52, its useful to override.


ThePhD:
Thanks for this. I should probably write some `require_script` and `require_file` examples at some point...

Nava2:
I also found out while doing this that you can not return anything other than a `table` from a modules. 

I'm so used to Javascript modules that I wasn't used to the Lua version.

PR #194, opened 2016-08-22
Add check for validity with object in `require_file`

### LINKED ISSUE #193: Return from require_file is not table (nil)
I can't seem to create a working version that causes this bug. 

For some reason, this section:
```C++
// state_view.hpp:54
optional<object> is_loaded_package(const std::string& key) {
			auto loaded = reg.traverse_get<optional<object>>("_LOADED", key); // nil if the file isn't loaded
			bool is53mod = loaded && !(loaded->is<bool>() && !loaded->as<bool>());
			if (is53mod) // <- always true
				return loaded;
```

This returns an `object` that stores `nil`. Regardless if the module has been previously loaded. 

Later, 

```C++
// stack_view.hpp: 95
template <typename Fx>
		object require_core(const std::string& key, Fx&& action, bool create_global = true) {
			optional<object> loaded = is_loaded_package(key);
			if (loaded) // should this be: if (loaded && loaded->valid())
				return std::move(*loaded);
			action();
			auto sr = stack::get<stack_reference>(L);
			if (create_global)
				set(key, sr);
			ensure_package(key, sr);
			return stack::pop<object>(L);
		}
```

This returned nil value (which is a valid `optional`) has the `nil` value moved into the result. Thus giving a bad value. 

Seems like line 98 should check if the result is a valid value not just that it's present.

Nava2:
Closes #193

### DIFF
--- a/sol/state_view.hpp
+++ b/sol/state_view.hpp
@@ -94,7 +94,7 @@ namespace sol {
 		template <typename Fx>
 		object require_core(const std::string& key, Fx&& action, bool create_global = true) {
 			optional<object> loaded = is_loaded_package(key);
-			if (loaded)
+			if (loaded && loaded->valid())
 				return std::move(*loaded);
 			action();
 			auto sr = stack::get<stack_reference>(L);


Nava2:
Back releasing this as 2.11.6 would be helpful or a new release.

Issue #203, opened 2016-08-30
Usertypes with many members become too complex for the compiler


OrfeasZ:
I'm experiencing this issue under MSVC140 (Update 3) on the 64-bit toolchain.

I have a usertype which has 147 property members (`sol::property`), a call constructor, and two base classes, and during compilation I'm getting the following errors:

```
2>I:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\type_traits(1016): fatal error C1202: recursive type or function dependency context too complex (compiling source file CustomUsertypeClass.cpp)
[...]

PR #225, opened 2016-09-17
Added helper functions for defining simple_usertype members manually.

OrfeasZ:
* Introduced a new simple_usertype class which derives from usertype.
* Exposed an "add" function for it in order to allow adding members one at a time before registration.
* Introduced state_view.create_simple_usertype<T>() methods for creating simple_usertypes.

As discussed in #203


OrfeasZ:
Note that gcc doesn't seem to like unqualified base classes. You might have to change [this](https://github.com/ThePhD/sol2/blob/68a9877a5e934316e583b78516bc20371a21e7f9/sol/usertype.hpp#L73) to read `: usertype<T>(...)` instead

PR #271, opened 2016-11-13
LuaJIT C++ Exception Full Interoperability Support & Missing ';' when SOL_NO_EXCEPTIONS is defined

aaiyer:
1. Missing ';' when SOL_NO_EXCEPTIONS is defined
2. If you are using a platform and compiler that has full c++ exception interoperability (http://luajit.org/extensions.html#exceptions), define SOL_LUAJIT_FULL_INTEROPERABILITY. This will prevent sol from catching (...) errors - in platforms & compilers than have full c++ exception interoperability Lua errors can be caught with catch (...) in C++ - in these cases sol inaccurately prevents Lua errors from being propagated correctly.

### DIFF
--- a/docs/source/exceptions.rst
+++ b/docs/source/exceptions.rst
@@ -45,9 +45,16 @@ If there is a place where a throw statement is called or a try/catch is used and
 LuaJIT and exceptions
 ---------------------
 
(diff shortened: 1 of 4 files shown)


ThePhD:
@aaiyer What do you mean by "incorrectly catches ( ... ) exceptions meant to be propogated" ? When you build Lua with C++ and then attempt to use `lua_error( ... );` ?

aaiyer:
On platform where LuaJIT supports Full interoperability -- "Lua errors can be caught on the C++ side with catch(...). The corresponding Lua error message can be retrieved from the Lua stack.". (See http://luajit.org/extensions.html#exceptions)

So, in these cases, Sol2 catches the error as "caught (...) exception" instead of calling lua_error(L).

ThePhD:
@aaiyer Thanks for the help here. I fixed up the docs you wrote and changed the name of it (since this define could technically apply to more than just LuaJIT). I also made a new release: https://github.com/ThePhD/sol2/releases/tag/v2.15.1

PR #301, opened 2016-12-16
fix thread_status::dead value for luajit

THE-FYP:
sol::thread_status::dead has an incorrect value when using sol with luajit. it occurs because luajit has no LUA_ERRGCMM and sol [adds its own](https://github.com/ThePhD/sol2/blob/develop/sol/compatibility/5.1.0.h#L51).

why
```cpp
enum class thread_status : int {
  ok = LUA_OK, // = 0
  yielded = LUA_YIELD, // = 1
  runtime = LUA_ERRRUN, // = 2
  memory = LUA_ERRMEM, // = 4 (it's ok)
  gc = LUA_ERRGCMM, // = 6
  handler = LUA_ERRERR, // = 5
  dead, // = 6! same value as thread_status::gc
};
```
i decided just assign thread_status::dead const value `-1` - it shouldn't cause any incompatibilities (i guess). anyway i don't insist this is the right solution so feel free to edit/discard the PR and fix this bug the right way.

### DIFF
--- a/sol/types.hpp
+++ b/sol/types.hpp
@@ -343,7 +343,7 @@ namespace sol {
 		memory = LUA_ERRMEM,
 		gc = LUA_ERRGCMM,
 		handler = LUA_ERRERR,
-		dead,
+		dead = -1,
 	};
 
 	enum class load_status : int {


ThePhD:
I did not notice that at all. I thought those values were incremental... or, well. Maybe they only were for 1 implementation of Lua.

`-1` is a good choice and shouldn't be used by Lua in the future or anytime soon (only positive numbers for that, IIRC).

PR #304, opened 2016-12-25
Correct the LuaJIT 2.1.0 condition

davedissian:
Should be LUAJIT_VERSION < 20100, not LUAJIT_VERSION <= 20100. Otherwise, it generates linker errors when LUAJIT_VERSION == 20100.

I also thought that the condition is clearer if written in the form `if (x < y)` rather than `if ((x-y) < 0)`

Tested and builds with no linker errors for LuaJIT 2.0.4 and 2.1.0 beta 2.


ThePhD:
Thank you!

PR #314, opened 2017-01-13
Fixed typo: luaxlib.h -> lauxlib.h

hurog:
This is totally nothing, but using 'SOL_USING_CXX_LUA' produced compile error.
Thank you for the great library!


ThePhD:
My bad. Thanks for submitting this and catching it.

Issue #324, opened 2017-01-26
Tests (custom-usertype) segfault when compiled with clang-cl


RaptorFactor:
When sol2 is built with clang-cl the "containers/custom-usertype" test segfaults during destruction of the state object.

Using latest clang-cl with latest MSVC. (Although results are the same with clang 3.9.1.)

```
clang version 4.0.0 (trunk)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Tools\LLVM-4.0.0-r291453-win64\bin
```

```
Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24215.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

Compiler Passes:
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\cl.exe:        Version 19.00.24215.1
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\c1.dll:        Version 19.00.24215.1
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\c1xx.dll:      Version 19.00.24215.1
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\c2.dll:        Version 19.00.24215.1
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\link.exe:      Version 14.00.24215.1
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\mspdb140.dll:  Version 14.00.24210.0
 C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\1033\clui.dll: Version 19.00.24215.1
```
Debug output:
```
0:000> g
(788.1950): Access violation - code c0000005 (first chance)
First chance exceptions are reported before any exception handling.
This exception may be expected and handled.
*** WARNING: Unable to verify checksum for sol2_tests.exe
sol2_tests!std::list<std::pair<const int, int>, std::allocator<std::pair<const int, int> > >::clear+0xdb:
00007ff7`ba5f34eb 488b00          mov     rax,qword ptr [rax] ds:dddddddd`dddddddd=????????????????
WARNING: Unable to find source file with matching checksum.  Found 'C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\list' with mismatch!
[...]

PR #339, opened 2017-02-18
Fix building tests with Catch v1.7.2.

RaptorFactor:
From the Catch docs:
Please note that the THROW family of assertions expects to be passed a single
expression, not a statement or series of statements. If you want to check a
more complicated sequence of operations, you can use a C++11 lambda function.

https://github.com/philsquared/Catch/blob/master/docs/assertions.md

In current versions of Catch violating this causes a build error.


RaptorFactor:
Tested on Windows under MSVC, GCC and Clang (Clang tests fail at runtime but that's still #324 and nothing to do with this).

Issue #331, opened 2017-02-03
Ability to search in documentation


traplol:
Unless I have been missing something, it would be very nice to be able to search for terms in the documentation.

I've been using sol2 for about a week now and it works great. My only problem has been when I read something in the documentation and at some point later I remember reading about that thing but I cannot remember _where_ in the docs that was.

traplol:
I just realized that http://sol2.readthedocs.io/en/latest/search.html exists, could that be made more visible or more accessible? For instance, put that at the top of each page of the documentation

ThePhD:
Yeah, I need to choose a better theme or customize it, but documentation customization is not my fortre.

ThePhD:
To anyone who's reading this: if you need a good Sphinx Theme I can choose that makes search easier, or have some information about how to customize a Sphinx theme, I could use that help.

ThePhD:
Praise be to @OrfeasZ !

PR #342, opened 2017-02-18
Added a search bar to the header of the documentation layout template.

OrfeasZ:
Implemented on top of the previously used "Haiku" theme as discussed in #331  
I personally tested this with Sphinx 1.5.2.

### DIFF
--- a/docs/source/_themes/sol2/layout.html
+++ b/docs/source/_themes/sol2/layout.html
@@ -1,0 +1,14 @@
+{%- extends "haiku/layout.html" %}
+
+{% block haikuheader %} 
+	<div class="header-left">
+		{{ super() }}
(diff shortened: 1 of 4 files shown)


OrfeasZ:
It looks like this:

squeek502:
This needs a bit of a tweak for paths outside of the root, like the `api` docs, which have a base URL of `http://sol2.readthedocs.io/en/latest/api/`, so when using the search form from one of the API pages it redirects to `http://sol2.readthedocs.io/en/latest/api/search.html` which doesn't exist (it should be `http://sol2.readthedocs.io/en/latest/search.html`).

OrfeasZ:
@squeek502 I'll take a look.

Issue #369, opened 2017-03-19
When searching documentation, as_table link leads to non-existent webpage


HoraceWeebler:
Go here:
http://sol2.readthedocs.io/en/latest/search.html?q=serial#

Click the top link for as_table.  It leads to a dead-end error page.

ThePhD:
Strange, it seems to append `.html` straight onto the original filename, without removing the `.rst`. I'll try to fix it soon.

ThePhD:
Oh boy, if I'm gonna fix how it outputs the titles I might have to go digging in this: http://www.sphinx-doc.org/en/stable/web/searchadapters.html

HoraceWeebler:
Maybe push the bug back on them?

squeek502:
Even stranger, this seems to only apply to specific search terms. Other terms work fine:

- http://sol2.readthedocs.io/en/latest/search.html?q=serialized#
- http://sol2.readthedocs.io/en/latest/search.html?q=test#
- http://sol2.readthedocs.io/en/latest/search.html?q=as_table#

OrfeasZ:
The issue seems to be the following: 

For some reason, RTD is using an old/unsupported version of the Sphinx search javascript file and instead of properly updating it, it replaces it with its own custom implementation which perform server-side calls for searching documentation (instead of the usual client-side lookups). There's a special edge case, where if the RTD server returns 0 results, it will fallback to the client-side search code. However, since that code is outdated, the results are not populated correctly.

I've submitted a hotfix with #370 which should (in theory) fix this issue, but keep in mind that if they decide to update the search scripts it will most likely break again. If that happens, then one simply needs to revert this commit.

ThePhD:
Fixed, thanks ot @OrfeasZ !

PR #370, opened 2017-03-20
Fixed broken links in search results when performing a local search

OrfeasZ:
This will **hopefully** work and address the issue in #369.


ThePhD:
Looks like black magic that I don't understand.

Approved 👍

Issue #196, opened 2016-08-23
Binding functions containing references to `std::vector<T>`


jnbrq:
Consider the following code:

```
#include <vector>
#include <sol.hpp>

int main(int argc, char **argv)
{
	typedef std::vector<double> V;
	sol::state lua;
	lua.new_usertype<V>(
		"Test",
		sol::constructors<
			sol::types<>,
			sol::types<std::size_t>,
			sol::types<const V&>
		>()
	);
	// lua["test_01"] = [](V &v) {}; // does not compile
	lua["test_02"] = [](V *v) {}; // compiles
	return 0;
}
````

When I try to register functions having references to `std::vector<double>` , it does not compile. The same thing also happens when trying to register them as member functions. No problem with constructors with references.

Happens on both Windows and Arch Linux.

for arch linux:
gcc version 6.1.1 20160802 (GCC)
lua 5.3.3

on Windows using MSYS2

This issue happens with the current version of single header version of the library.

With some previous version, it didn't happen. I do not remember the version of that revision, so I upload the header to ubuntu pastebin:

[Ubuntu pastebin link](http://paste.ubuntu.com/23081218/)

ThePhD:
This is a known regression (known as of a little bit ago, but known).

It was the primary problem with serializing vector -> table upon push, and allowing deserializing from a table as well. I bit the bullet and added it, but now I'm in... well, this conundrum!

I'm going to have to put a fair amount of work in to make this behave proper.

ThePhD:
Better handled now, try the latest: https://github.com/ThePhD/sol2/releases/tag/v2.12.0

jnbrq:
Thank you :+1:

Nava2:
Not sure if it was considered, but sol fails to properly bind `std::set`.

ThePhD:
Yeah, `std::set` was ignored because it's a speshul snowflake. I feel like I'd have to write an explicit overload just for it.

ThePhD:
Particularly: how do you make a table or a table-alike for `std::set`? It only contains keys, not values, so the table structure of Lua does not fit its mental model. Furthermore, even if we DID try to force it to work (assuming we banned the syntax `mytable["key"] = value;` from being legal), what does `local val = mytable["key"];` return? `true` if it exists, `false` if it doesn't? That's... really not a useful thing to have, honestly.

ThePhD:
Backported because WHY NOT

[...]

Issue #189, opened 2016-08-22
HELP - sol2 Users


ThePhD:
Hello, users of sol2! This issue was made so that I could get a better idea of what you all use for Lua and how much you like / dislike sol2 compared to what you used to use (if anything) and sol2 now. If anyone could just reply to this issue with their thoughts, perhaps answer some of the questions below (or go on a tangent, really anything), I'd appreciate it. I'm going to be writing a paper about sol2 and the Lua landscape, so I'd like to get to know what my users use sol2 for:

- How did you find out / hear about sol2?
- What Lua library did you use before you found sol2?
- What do you like about sol2?
- What do you dislike about sol2?
- Is it easier or harder from the last Lua library you used? (If you did use a library or the plain C API before this)
- Is the [documentation](http://sol2.readthedocs.io/en/latest/) or (examples)[https://github.com/ThePhD/sol2/tree/develop/examples} helpful?
- Anything else about the library you feel like adding or pointing out?

Thanks for much for reading and considering, and double thanks if you do reply!

Nava2:
> * How did you find out / hear about sol2?

Was investigating `sol` originally, but found there to be bugs which were fixed in a `sol2` developer branch. 

> * What Lua library did you use before you found sol2?

Did not use any extensively, however, I was investigating `sol` and `luacppinterface`. 

> * What do you like about sol2?

Syntax is much cleaner than other libraries. 

> * What do you dislike about sol2?

Take the following with a grain of salt, as these are negatives, but I also understand why they exist. 

* Compile times: When creating even simple structures, the compiler takes a world of time to figure out loading
* Error messages: The current error messages when a compiler error is made are opaque and often are *not* an indication of the user error because they propagate so deeply before surfacing.
* Build tool: As a multi-platform developer, it would be helpful to have the project building with a classical build tool like `cmake` for integration purposes

> * Is it easier or harder from the last Lua library you used? (If you did use a library or the plain C API before this)

N/A

> * Is the documentation or (examples)[https://github.com/ThePhD/sol2/tree/develop/examples] helpful?

The documentation has become excellent and you're great at improving it when brought to your attention. 

> * Anything else about the library you feel like adding or pointing out?

Keep up the activity, it's great to see. What would be helpful is having a means of "asking for help." Github issues don't seem like the place for that and there's a lot of people asking questions.

Grant1219:
> How did you find out / hear about sol2?

[...]

PR #402, opened 2017-05-03
Add :find to containers to make std::[unordered_]set more useful Lua-side

carloscm:
std::[unordered_]set had no way to perform a lookup from the Lua side (other than iterating), and I saw in #196 it wasn't desired to have table-like access for lookups, so I added a :find to the container metatable.

### DIFF
--- a/docs/source/api/containers.rst
+++ b/docs/source/api/containers.rst
@@ -76,7 +76,7 @@ Based on the type pushed, a few additional functions are added as "member functi
 * ``my_container:clear()``: This will call the underlying containers ``clear`` function.
 * ``my_container:add( key, value )`` or ``my_container:add( value )``: this will add to the end of the container, or if it is an associative or ordered container, simply put in an expected key-value pair into it.
 * ``my_contaner:insert( where, value )`` or ``my_contaner:insert( key, value )``: similar to add, but it only takes two arguments. In the case of ``std::vector`` and the like, the first argument is a ``where`` integer index. The second argument is the value. For associative containers, a key and value argument are expected.
-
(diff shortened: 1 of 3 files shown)


squeek502:
Might be worth mentioning in the docs that it differs from the C++ `find` with what gets returned (when the value is found, it returns the value, and when the value is not found, it returns `nil`)?

ThePhD:
Looks good. Thanks for including this: I hadn't thought of it when using `set`.

Finally, as to the comment you left in #189 -- we can't use that method for looking things up (as far as my understanding goes). If you serialize all functions into the C++ usertype, we automatically do a no-strings lookup that calls into the C++ function through a Lua table. This gets the speed you mention. However, it is impossible for that call type to work with property syntax and still be hooked up to C++ variables (`obj.something = 2` and `val = obj.something`), because of the way __index and __newindex work and how they propogate the failure table when a lookup goes bad (last failed table/userdata, not the one that started the original lookup failure in the first place).

If you can find out a scheme that makes all tests pass, though, please feel free to hack on it.

carloscm:
One thing I forgot to mention, I changed has_find in my patch, since the previous one wanted to match find() as opposed to find(C::value_type -alike), as far as my C++ metaprogramming-fu can understand. Nothing appears to be broken after change tho.

@squeek502 I assumed that since an iterator return is kind of nonsensical on the Lua side (or for a set, my use case), it was going to be for sure value + Lua convention of nil on not found, so there wasn't much to comment about the return types.

@ThePhD Indeed it would require all the fields, both for the class and all the base ones, to be in the table/vector for the first-chance lookup. So maybe it wouldn't work for all possible cases.

PR #415, opened 2017-05-30
Added comparison operator to reference

haikarainen:
... to allow for sol::functions to be used as a key in `std::set`, `std::map` etc containers.

Ugly hack, and may be possible to do it more properly (to make it actually ordered!), but this should work good enough in a `std::set` for example

### DIFF
--- a/sol/reference.hpp
+++ b/sol/reference.hpp
@@ -224,6 +224,10 @@ namespace sol {
 		return lua_compare(l.lua_state(), -1, -2, LUA_OPEQ) == 1;
 	}
 
+	inline bool operator< (const reference& l, const reference& r) {
+		return !(l == r);
+	}
+
 	inline bool operator!= (const reference& l, const reference& r) {
 		return !operator==(l, r);
 	}


mrgreywater:
Would probably make more sense to use lua_topointer to get the pointer to the functions and compare those. The 'less than' operator needs the following statement to be true, which is not the case with your implementation:
´´´
r < l == (!(l < r) && l != r)
´´´

ThePhD:
This PR breaks down immensely when a `sol::reference` is used as a key type.

The proper thing to do is to define your own `less_than_comparator` function object and hand it to `std::set` or `std::map` to use (see [the 'Compare' type argument to std::set here](http://en.cppreference.com/w/cpp/container/set)). Then you can define your own logic specific to your needs.

There is no standard comparator for these types which is either useful or helpful to the user in the end, and while I appreciate the PR I cannot integrate this functionality into sol2 and promise to support it when it is at best a hack for a single use case.

haikarainen:
Yeah I agree, this was shabby. Thanks for the info though! :)

Issue #422, opened 2017-06-13
(lacking) type safety of T -> unique_usertype<T>


rohlem:
(Preface: Yo. Sol2 is amazing. Many thanks.)

After some initial testing with unique_usertype types I came to the conclusion that Sol wholly equates a unique_usertype to its `::type` proxee, the type it represents.
(My best attempt at) a minimal test scenario:
```
//SOL_CHECK_ARGUMENTS isn't of any help either
#define SOL_CHECK_ARGUMENTS
#include <Sol/sol.hpp>

#include <iostream>
#include <memory>

// make a type
struct test{int special_value = 17;}; //never reset, should only ever be 17
// test function that takes a unique_ptr argument
void print_up_test(std::unique_ptr<test>& x){
	std::cout
		<< static_cast<void*>(x.get()) << " : " << std::flush
		<< x->special_value << std::endl;
}

int main(){
	// setup ...
	sol::state s;
	s.open_libraries();
	// tell Lua about the function
	s.set_function("print_up_test", &print_up_test);
	// unique_ptr object works as expected
	s["up_test"] = std::make_unique<test>();
	s.script("print_up_test(up_test)");
	// normal object is blindly accepted, leading to a bogus address with either a garbage value or triggering a segfault
	s["direct_test"] = test{};
	s.script("print_up_test(direct_test)");
}
```
Output on my setup (some mingw-w64 g++ 7.1.0):
```
0x339740 : 17
0x8800002b4f7b1624 : <program crash>
```
The same behaviour is observed when calling a functor created by sol::overload. (tested in C++17 as well as C++14 mode, just in case)

Maybe this behaviour is by design, I couldn't find exact wording regarding the type conversions in the documentation however (other than the fact that unique_ptr -> pointer works without issue), so I was quite surprised of this silent error. Specializing sol::is_unique_usertype of the specific unique_ptr specialization to std::false_type removes the implicit illegal cast, rejects the argument in the above example with SOL_CHECK_ARGUMENTS and makes use cases with sol::overload behave correctly/usefully.

Many applications will probably differentiate the usage of Lua owned objects and proxies, so it would be understandable to leave this unsupported/undefined; my original use case would have involved a bunch of custom iterators/proxies of mixed-owning nature, but making them non-unique_usertype and putting `:get()` calls in between in Lua doesn't pose a huge usability issue either. :P

ThePhD:
[...]

PR #423, opened 2017-06-14
fix usage of unique_ and shared_ptr as non-unique-usertype

rohlem:
This PR allows std::unique_ptr and std::shared_ptr to be used as usertypes (if the unique_usertype trait for the specific specialization has been disabled, I didn't put much thought into whether it would make any sense otherwise).
Includes a test case (leads to compilation errors on develop), a 100% safe fix for the issue itself (incorrect detail::deref overload) and a 99% safe workaround for the new usage of detail::deref; see the (surprisingly small) commits for details.
(P.S.: The test suite is _really_ easy to set up if you just `pip install ninja` and play with include/library directories/flags. Cool stuff.)

### DIFF
--- a/sol/traits.hpp
+++ b/sol/traits.hpp
@@ -407,6 +407,16 @@ namespace sol {
 			return *item;
 		}
 
+		template <typename T>
+		inline const T& deref(const T& item) {
+			return item;
+		}
+
+		template <typename T>
+		inline T& deref(T& item) {
+			return item;
+		}
+
 		template<typename T, typename Dx>
 		inline std::add_lvalue_reference_t<T> deref(std::unique_ptr<T, Dx>& item) {
 			return *item;
--- a/sol/usertype_metatable.hpp
+++ b/sol/usertype_metatable.hpp
@@ -413,7 +413,7 @@ namespace sol {
 					}
 					else {
 						Op op;
-						return stack::push(L, (detail::ptr(l) == detail::ptr(r)) || op(detail::deref(l), detail::deref(r)));
+						return stack::push(L, (detail::ptr(l) == detail::ptr(r)) || op(detail::deref<T>(l), detail::deref<T>(r)));
 					}
 				}
 			}
--- a/test_usertypes.cpp
+++ b/test_usertypes.cpp
@@ -265,6 +265,13 @@ struct matrix_xi {
 	}
 };
 
+struct type_without_unique_usertypes {};
+
+namespace sol{
+	template <> struct is_unique_usertype<std::unique_ptr<type_without_unique_usertypes>> : std::false_type {};
+	template <> struct is_unique_usertype<std::shared_ptr<type_without_unique_usertypes>> : std::false_type {};
+}
+
 TEST_CASE("usertype/usertype", "Show that we can create classes from usertype and use them") {
 	sol::state lua;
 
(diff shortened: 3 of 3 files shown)


rohlem:
just realized I forgot to rebuild the single-header sol.hpp ... though I didn't see it in other PRs. Is it a part of your build bot or something?

ThePhD:
Well, the test fails for a specific compiler on LuaJIT, so I'm going to have to look into this.

Finally, I don't quite understand the use-case here. `std::unique_ptr` is a decorator that conveys ownership, and `std::shared_ptr` is the same. It does not actually change how the internal `T`'s methods and calls work. What is the purpose of allowing this specialization in particular?

rohlem:
The test profile is exactly the same as since the last commit (written by you, just initiated by my issue #422 ), runtime-failing in the test case also introduced in the same commit. Compare:

- https://travis-ci.org/ThePhD/sol2/builds/242582374?utm_source=github_status&utm_medium=notification

- https://travis-ci.org/ThePhD/sol2/builds/242870689?utm_source=github_status&utm_medium=notification

both fail at test_functions.cpp:1139 because of an exception (i.e. there's something wrong with the functionality just introduced, being tested in that test case...) .

For one it's a workaround for distinguishing different usertype proxies (as mentioned in #422 ), by not making them unique_usertypes but instead making each their own usertype.

Then you might want to put all lifetime managing operations into the metatables of owning constructs (unique_ptr, other iterators/proxies, just not the type itself with the current design) 's metatables (f.e. "destroy" -> up.reset(), it->container.remove(it) etc.) and making the reference metatable (of T*) observer-only.

If anything, the current design treats unique_ptr and shared_ptr specially (IMHO by error, the overload resolution is clearly non-intuitive in this case) and it might seem unintuitive for users that new_usertype<std::unique_ptr<ANY_T>> always leads to errors and cannot be made to work under any circumstances. Without this PR, that is.

ThePhD:
Your test cases don't do anything other than test the ability to compile. It doesn't convey what semantics should be present: you gain the ability to differentiate a single type from the rest of its containing types in a `sol::overload` context, but what happens after that?

Can you call a lua function with a unique_ptr of this type and just take a `T*` or `T&`? Does modifying the `unique_ptr` propagate its changes properly?

The overload resolution behavior may not be what you want intuitively, but there's other considerations to make in this runtime system that you are not considering with this fix. If I were to change the current behavior, I would commit more fully to finding a low-overhead or no-overhead (space, and time) way to identify what a specific type unique-like type `T` is that still maintains a lot of the invariants that sol2 users have not only grown accustomed to but they have found useful for quite some time now.

I have a few ideas for how to fix the issue you're having in a more complete manner that takes no extra space and costs very little time (only 1 pointer comparison). I believe it will solve this problem in a much more elegant manner without having to selectively turn off the `unique_usertype` behavior and trigger failovers on functions that take a `T*` or `T&`.

If you want to keep working on this PR or find some other way to enable what you're looking for, I'll be happy to compare the final version of this to the final version of what I come up with, but as it stands I'm not sure that I can accept this as it is.

ThePhD:
I have finished my commit that should solve your overload problem (as in #422) without having to have to opt-out of `unique_usertype` semantics. It provides 0 overhead and comprehensive testing shows that it does exactly what's needed (though we have to see if LuaJIT still barfs on the tests).

Let me know if this is preferable or if you still need to opt-out of `unique_usertype` still for some other reason.

ThePhD:
Since this is solved by my latest, I'm going to close this.

rohlem:
(I'm so sorry it took me so long to respond, I thought I'd get to it sooner.)
Your commit indeed fixes the overload issue, and so my last argument for allowing unique_ptr and shared_ptr to be usertypes of their own is shattered. Gz to pull it off after all, and thank you for your continued efforts.

Trying to register unique_ptr as a usertype still leads to confusing compile errors, so I'd suggest adding a static_assert like
  `static_assert(not sol::is_unique_usertype<T>::value, "Do not register a unique_usertype as usertype! Register its base type instead!);`
(and the same for unique_ptr<T> and shared_ptr<T> specifically) in the right places might help users to find the issue.
Just a suggestion though, I won't create another issue/PR myself; I feel like I've already annoyed you enough.
Thank you again. ^^

Issue #414, opened 2017-05-30
Fail to compile on gcc 7.1.1


Strasser-Pablo:
Hi,
I just updated my gcc version and am unable to compile the sol2 code, I guess it is because of change introduced in c++17.
Here is an excerpt of the error:

 [1/48] compiling test_storage.cpp to obj/test_storage.o
FAILED: obj/test_storage.o 
g++ -MMD -MF obj/test_storage.o.d -c -Wall -Wextra -Wpedantic -pedantic -pedantic-errors -std=c++14 -ftemplate-depth=1024 -DNDEBUG -O3 -I"." -I"./include" -I"/usr/include/lua5.3" -I"/usr/local/include/lua5.3" -isystem"Catch/include" -Werror test_storage.cpp -o obj/test_storage.o
In file included from ./sol/stack.hpp:27:0,
                 from ./sol/object.hpp:26,
                 from ./sol/proxy.hpp:26,
                 from ./sol/table_core.hpp:25,
                 from ./sol/table.hpp:25,
                 from ./sol/state_view.hpp:26,
                 from ./sol/state.hpp:25,
                 from ./sol.hpp:45,
                 from test_storage.cpp:4:
./sol/stack_check.hpp: In instantiation of ‘static bool sol::stack::checker<T, (sol::type)6, C>::check(lua_State*, int, Handler&&, sol::stack::record&) [with Handler = int (&)(lua_State*, int, sol::type, sol::type) noexcept; T = sol::coroutine; C = void; lua_State = lua_State]’:
./sol/stack_core.hpp:233:69:   required from ‘bool sol::stack::check(lua_State*, int, Handler&&, sol::stack::record&) [with T = sol::coroutine; Handler = int (&)(lua_State*, int, sol::type, sol::type) noexcept; lua_State = lua_State]’
./sol/stack_core.hpp:239:19:   required from ‘bool sol::stack::check(lua_State*, int, Handler&&) [with T = sol::coroutine; Handler = int (&)(lua_State*, int, sol::type, sol::type); lua_State = lua_State]’
./sol/coroutine.hpp:87:48:   required from here
[...]

PR #424, opened 2017-06-14
noexcept function c++17 hotfix

rohlem:
This PR extends the template trait struct system for functions in meta and meta_detail (basic_traits and member_function_wrapper) to fix noexcept functions in C++17. Comes with two test cases (one for usertypes, just in case).

I assume you have some greater plan moving to C++17, and there might be some more improvements in some other areas, but in my usage of Sol this is the only thing that ever came up and it looked easy to do, so here it is. All changes are hidden behing preprocessor if-s with the correct feature-testing macro, so there should be no regressions (or at all notable change really) in any C++14 builds.

(old info, now commit 3 is removed: Additionally, the last commit changes 3 places in which the tests and examples use the return type conversion to string mentioned in issue #414 , just to make it actually buildable (on my setup, mingw-g++ 7.1.0). Feel free to omit that commit (or tell me to remove it, if that's easier) if you'd prefer dealing with those spots yourself.)

### DIFF
--- a/sol/bind_traits.hpp
+++ b/sol/bind_traits.hpp
@@ -54,12 +54,13 @@ namespace sol {
 			template <std::size_t I, typename T>
 			using void_tuple_element_t = typename void_tuple_element<I, T>::type;
 
-			template <bool has_c_variadic, typename T, typename R, typename... Args>
+			template <bool it_is_noexcept, bool has_c_variadic, typename T, typename R, typename... Args>
 			struct basic_traits {
 			private:
 				typedef std::conditional_t<std::is_void<T>::value, int, T>& first_type;
 
 			public:
+				static const bool is_noexcept = it_is_noexcept;
 				static const bool is_member_function = std::is_void<T>::value;
 				static const bool has_c_var_arg = has_c_variadic;
 				static const std::size_t arity = sizeof...(Args);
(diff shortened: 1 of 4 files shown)


ThePhD:
The code here is g++/clang++ specific and does not consider VC++. I will need to vet this code and everything about it before accepting a PR.

rohlem:
The very last commit is a fix for g++ and clang++s current C++17 behaviour, it sacrifices the readability of the example and test code but otherwise cannot lead to any regressions of anything. If you don't want it I'll gladly remove it from this PR, I just thought it helps to be able to build it (otherwise C++17 builds are broken in g++ and clang++).

The preprocessor-ifs are SD-6 Feature Test macros according to [gcc's documentation](https://gcc.gnu.org/projects/cxx-status.html) , so they come from standards people and we'd hope that implementers correctly specify them.
Undefined macros evaluate to 0 iirc, so the ifs lead to false and there is no change to behaviour.
If you want more thorough/distinct testing a different macro in some compatibility header file would be the way to go (right?) and this standardized constant is a good way to start.
Either way, compliant C++14 behaviour is unaffected.

rohlem:
ok, scratch that last commit, all the build bots (none of which use gcc 7) fail with explicit static_casts. I cannot understand how an implicit conversion changes any of that, but I'll remove that commit from this PR immediately, since it's clearly not working out.

Nava2:
@ThePhD have you considered [AppVeyor](https://www.appveyor.com/docs/build-environment/#visual-studio-2017) as they have VS 2017 support?

ThePhD:
@Nava2 I have, but I just haven't had the time to get to do it. I was planning to do it when I started adding test harnesses for g++ 5, 6, and 7 | clang++ 3.7, 3.8, 3.9 and 4.0 while putting all the older ones is "allowed failures" buckets. I also wanted to look into getting more updated OSX images on travis-ci so I could finally see if I could get `std::tuple`'s compiler bug that's been plaguing builds from old XCode images to be corrected...

That's going to take a while though.

ThePhD:
A quick check to the VS2017 feature table reveals that `noexcept` is not part of the type system just yet: https://blogs.msdn.microsoft.com/vcblog/2017/05/10/c17-features-in-vs-2017-3/

rohlem:
(I'm so sorry it took me so long to respond, I thought I'd get to it sooner.)
Thanks for taking it into your own hands, and happy to see there's potential for improvements in a lot of places; I hope I didn't slow you down too much. Keep it up! ^^

PR #441, opened 2017-07-02
#include fixes

rohlem:
Just some inclusion order fixes I came across that prevented selective inclusion.
(I didn't go out of my way to test _every_ header... though at some point I'd totally be up to do that too if you want me to.)
In case you want to verify, some affected headers were `"sol/object.hpp"`, `"sol/table_core.hpp"`, `"sol/usertype_metatable.hpp"` and `"sol/coroutine.hpp"`.

The second commit might be more interesting, previously the `SOL_SAFE_FUNCTIONS` macro was broken because `function.hpp` used the `sol::function`-alias which would resolve to `protected_function`, undefined/incomplete at that point. So I just conditionally added `#include "protected_function.hpp"`... should be safe, right?

Also I rebuilt the single header (your script btw doesn't normalize line endings, but I did that manually), but because the ordering changed that last commit is huge, so feel free to only accept the first two or something.

### DIFF
--- a/single/sol/sol.hpp
+++ b/single/sol/sol.hpp
@@ -20,8 +20,8 @@
 // CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 // This file was generated with a script.
-// Generated 2017-07-04 08:40:46.845523 UTC
-// This header was generated with sol v2.18.0 (revision 345a398)
+// Generated 2017-07-07 04:42:49.277472 UTC
+// This header was generated with sol v2.18.0 (revision 0287c6e)
 // https://github.com/ThePhD/sol2
 
 #ifndef SOL_SINGLE_INCLUDE_HPP
@@ -31,29 +31,6 @@
 
 #ifndef SOL_HPP
 #define SOL_HPP
-
-#if defined(UE_BUILD_DEBUG) || defined(UE_BUILD_DEVELOPMENT) || defined(UE_BUILD_TEST) || defined(UE_BUILD_SHIPPING) || defined(UE_SERVER)
-#define SOL_INSIDE_UNREAL
-#endif // Unreal Engine 4 bullshit
-
-#ifdef SOL_INSIDE_UNREAL
-#ifdef check
-#define SOL_INSIDE_UNREAL_REMOVED_CHECK
-#undef check
-#endif 
-#endif // Unreal Engine 4 Bullshit
-
-#ifdef __GNUC__
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wshadow"
-#pragma GCC diagnostic ignored "-Wconversion"
-#if __GNUC__ > 6
-#pragma GCC diagnostic ignored "-Wnoexcept-type"
-#endif
-#elif defined _MSC_VER
-#pragma warning( push )
-#pragma warning( disable : 4324 ) // structure was padded due to alignment specifier
-#endif // g++
 
 // beginning of sol/state.hpp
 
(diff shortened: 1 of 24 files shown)


rohlem:
oh whaaaaat I forgot to check the test suite, give me an hour or so

ThePhD:
I don't really expect sol2 to be selectively included, but vetting the framework might be nice...?

I also should probably make a `forward.hpp` containing forward declarations for every type in sol2...

rohlem:
uugghh I messed up twice. And the commit message says usertype when I meant userdata.
Whatever, now it works. Checking all other headers now that I'm at it.

rohlem:
Alright, I think it's done now.
It actually wasn't as much work as I'd anticipated, most other headers worked out well.

g++ generates some warnings however, which are currently fixed by macros in sol.hpp, but still trigger with selective inclusion. The last three commits address this:
- I moved the macros before and after the includes in sol.hpp into files `file_begin.hpp` and `file_end.hpp` with appropriate include guards to be used for every sol header in succession.
- To keep the behaviour the same I added "override" files `global_begin.hpp` and `global_end.hpp` to be placed into `sol.hpp` directly.
- Then I added `file_begin.hpp` and `file_end.hpp` includes into all relevant files (the files leading to warnings on g++ and optional_implementaiton.hpp, the only place I could find alignment).
- It wouldn't hurt to put them in every other file as well, but I don't think it's necessary and you can adopt as you go / problems arise.

Not sure whether I overdid it... but now that it's done it doesn't strike me as _too_ ugly personally... though I'll of course remove it again if you don't like it.

rohlem:
Well, clang certainly didn't like it... I've now tried fixing it, if it still doesn't work you'll have to help me out .-. (or just tell me to revert those commits)

rohlem:
Derp, didn't read ahead. Apparently clang accepts both `#pragma GCC diagnostic` and `#pragma clang diagnostic` ? Changed to `clang` to be safe.

rohlem:
Rebase (=> force push) + small fix, apparently I misread an error message when fixing the test suite earlier so "object.hpp" was broken again; now everything works again.

rohlem:
Clarification of what I tried to do:

`sol.hpp` has a bunch of macros at the beginning and at the end of the code. The way the stack-based warning/"diagnostic" system works it is undesirable to have user code in between this pair of chunks. Therefore, to support selective inclusion, every (relevant) header will need to include both the beginning and the ending chunk.
To avoid copy-pasting them around the code base I put them into the files `file_begin.hpp` and `file_end.hpp`. The only reason they don't have traditional include guards is because they have to trigger several times in a row, for every header that includes them; `file_end.hpp` "resets" `file_begin.hpp` 's "include guard" (and emits an error if it wasn't defined yet).
I cannot think of an easier way to package the current macros on a per-file basis; note that users might include the files in any order. "Leaking" the diagnostic settings might silence errors in user code, omitting the macros means we don't silence sol's internal warnings.

`global_begin.hpp` and `global_end.hpp` are just an "optimization" of sorts; when including `sol.hpp` (or the single-header), instead of having several file_begin and file_end sections cluttered throughout, `global_begin.hpp` includes `file_begin.hpp` once at the very top and "disables" all further inclusions until `global_end.hpp` undefines its macro and includes `file_end.hpp`, again, only once.
That step was done purely to reduce the generated/effective code size; it is in no way necessary and simply removing all mentions of `global_begin.hpp` and `global_end.hpp` leads to no change in behaviour (as long as `file_begin.hpp` and `file_end.hpp` are in every file they need to be).

rohlem:
I added commits that remove `global_begin.hpp` and `global_end.hpp`, hopefully `file_begin.hpp` and `file_end.hpp` are concise enough.
(Otherwise you could remove those too, but that currently means selective inclusion always triggers warnings, and future macro inconsistencies, since including `sol.hpp` afterwards means the headers included earlier missed out on those directives.)

ThePhD:
Okay, I understand that it's for selective inclusion. But _why_ do you need selective inclusion? Is there a compilation speed benefit to including only parts of sol2? Note that almost everything gets used by `sol::state`, so selective inclusion doesn't even make sense here except if you had your own `lua_State*` and were working with the stack directly.

rohlem:
Well, yes, the primary motivation was a supposed compilation speed improvement. I didn't benchmark it, but there's definitely the _opportunity_ for quite some time save when f.e. using sol types in a function prototype of an often-included header central to the project.

You are correct in that sol has a tightly-woven structure, two examples for you to judge:

- `object.hpp`  eventually includes both `function.hpp` and `stack.hpp` (while `function.hpp` does not depend on `object.hpp`), totalling in 43/70 headers (which is admittedly more than I would have expected before counting just now);  However none of those headers include `state_view.hpp` or `state.hpp`.

- `types.hpp`, sufficient for most declarations and user-side trait specializations for sol types, only ever needs 11/70 headers, avoiding all implementation details of `stack.hpp` etc. .

All of that is probably only important to the collective OCD however, so two (possibly) more rational aspects:

- Sol includes a lot of templating utilities (type name demangling, function property deduction, its own portable `optional` type) as well as some general abstractions that are useful to have and re-use around the project (especially when just starting out, but I'd argue also as a permanent solution if hidden behind typedefs/using-declaration to replace in case sol's implementation eventually doesn't cut it).

- It's, from my limited experience, the obvious and idiomatic standard for (tidy and well-maintained) C++ libraries not to break horribly when trying to selectively include them. If you really don't want to deal with that it should be prominently mentioned somewhere in the documentation (though if it was I'd have, admittedly, probably ignored it at first...) .

That's all just my opinion/suggestion however, feel free to request further changes (removing `file_*.hpp`) or cherry pick from this PR to your liking.
I guess seeing it as sheer novelty that only ever came up 2+ years into people using the library it probably does seem rather pointless. It probably is.
You're the expert, obviously. (8 And thanks again for everything.

ThePhD:
Let's assume we want selective inclusion.

One of the things we would need to define is what headers include what bits of code. For example, the documentation is written as if you are getting all of sol2: what happens when selective inclusion is a supported scenario?

How does somebody include types and functions that honestly have no other purpose but to serve as utility in the framework, but are expected to be there (`as_args`, `as_returns`, `variadic_args`, `variadic_results`, `this_environment`, ... )? What does the header situation look like at that point?

I'm all for faster compile-times, but I would also need to consider changing sol2's project structure as well. Do I put 4 or 5 designated headers that give singular units of functionality in the top-level and shove everything else in a `sol/detail/` folder? Or are they allowed to include on a per-file granularity? When does it become okay to bring the 5, 6 utility headers along, and when is it okay to drop them like a hot sack of potatoes? Or, do I make the user include `as_args`, `as_returns`, ... by themselves, and offer a top-level grouping header?

These are the questions that need answers before I commit to this.

ThePhD:
I'm going to close this, since I don't think it's needed.

As an addendum, I'm adding and fleshing out a `sol/foward.hpp`, which may alleviate some of the concerns.

PR #446, opened 2017-07-06
enable move semantics when pushing sol::optional<T>

rohlem:
A rather straight-forward fix; `sol::optional<T>`'s `stack::pusher` specialization takes the `optional` by forwarding reference but then simply pushes `t.value()`.
This commit adds a `static_cast` that correctly forwards the contained value according to the argument's reference type. (A general `forward_like` utility function might be handy, but I wasn't sure where to put it and if it's the only use case then it doesn't really matter yet.)
Test case (move-only struct, a `sol::optional` of which cannot be pushed before the fix) included (in `test_tables.cpp`, because I couldn't think of a better place to put it).

### DIFF
--- a/sol/stack_push.hpp
+++ b/sol/stack_push.hpp
@@ -26,6 +26,7 @@
 #include "raii.hpp"
 #include "optional.hpp"
 #include <memory>
+#include <type_traits>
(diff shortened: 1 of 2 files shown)


ThePhD:
Thanks for the help.

Issue #470, opened 2017-08-04
incorrect handling of large 32bit unsigned integers and 64bit signed integers


mrgreywater:
Platform: MSVC x86 (luajit 2.1b3)
```
    lua.set_function("largeinteger", []()->std::uint32_t {
        return 0xFEFEFEFE;
    });
```
when calling this function from lua, you'll get a negative number, because 32bit unsigned integers and 64bit integers are pushed as lua_Integer (32bit signed)

See https://github.com/ThePhD/sol2/blob/e69e7c79faaa1f91b6cc736a4b08419a4da3219b/sol/stack_push.hpp#L194#L218

I haven't checked the other way around, passing large lua numbers to sol function, but it may show similar behaviour.

ThePhD:
What exactly becomes the proper way, then, to handle these integers? I could try to push a `double` for 32-bit integers, but this would be unacceptable when we moved up to 64-bit integers since a double couldn't represent their largest values (signed or unsigned) anyhow.

mrgreywater:
Well a double has 53 significant bite to represent the integer. I would expect the exact lua number representation for 32bit and the closest approximation for 64bit values. Not suddenly a forced 32bit signed integer. Casting to double first would work best.

ThePhD:
As a side note, this would also mess with 5.3 code where there's a clear distinction between `int` and `double`, at least when you use API-checking functions to determine which is which.

ThePhD:
As another additional sidenote, if you build Lua as a 64-bit library it defaults to using `long long`, which will alleviate this problem somewhat as well. Note to self: purposefully build in 32-bit to make sure that these classes of errors don't go over head...

demianmnave:
At least 5.2 supports the notion of unsigned integers  (e.g. `lua_pushunsigned` and `lua_tounsigned{x}`).  Aren't these functions sufficient to address the problem regardless of word size?

ThePhD:
I believe that was part of the reason why I had to change it: supporting 5.1 meant backtracking to using older functions. I could do some version-sniffing to protect against this, albeit 5.1 users will still flounder.

demianmnave:
The version sniffing could be localized to  `sol_pushunsigned` and `sol_tounsigned{x}`, for example.  This would likely help avoid interfering with default Lua behavior, as well.

I'm probably preaching to the choir here, but having Sol2 deal with unsigned integers via double seems problematic for a variety of reasons, not the least of which is the potential to violate the principle of least surprise.

[...]

PR #476, opened 2017-08-07
Large integer support

mrgreywater:
This PR adds large integer support for x86 and integer misrepresentation detection for x86/x64.
sol2 will still try to use lua_Integer wherever it is safe to do so, only when the value can't possibly fit in lua_Integer, it uses lua_Number as fallback with optional runtime checks to make sure the floating point value still represents the integer accurately.

Would appreciate any feedback or concerns, if there are any.

See #470

- [x] large integer changes
- [x] tests
- [x] working linux:i386 build
- [x] passes travis-ci (as good as _develop_)
- [ ] reviewed

unrelated to this PR: 
- The build for the original develop branch is currently broken, I'll update my fork from upstream as soon as it's fixed.
- currently both linux:i386 and windows:x86 are failing some tests on upstream/develop (at least with the latest luajit) maybe we could have some i386 builds on travis?
- you manually need to set SOL_EXCEPTIONS_SAFE_PROPAGATION under windows x86/luajit2.1b3 (the new versions have proper exception handling under windows x86 with \EHsc)
- single.py is failing with python2.7 and only works with python3
- it would be nice to have a build config that runs with msvc that you don't need to create yourself

### DIFF
--- a/single/sol/sol.hpp
+++ b/single/sol/sol.hpp
@@ -20,8 +20,8 @@
 // CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 // This file was generated with a script.
-// Generated 2017-08-11 14:35:51.568053 UTC
-// This header was generated with sol v2.18.0 (revision 4f7f1af)
+// Generated 2017-08-11 12:59:23.551177 UTC
+// This header was generated with sol v2.18.0 (revision 5e109c2)
 // https://github.com/ThePhD/sol2
 
 #ifndef SOL_SINGLE_INCLUDE_HPP
@@ -5798,6 +5798,7 @@ namespace sol {
 
 // end of sol/inheritance.hpp
 
+#include <cmath>
 #ifdef SOL_CXX17_FEATURES
 #endif // C++17
 
@@ -5854,7 +5855,14 @@ namespace sol {
 			template <typename Handler>
 			static bool check(lua_State* L, int index, Handler&& handler, record& tracking) {
 				tracking.use(1);
-				bool success = lua_isinteger(L, index) == 1;
+#if SOL_LUA_VERSION >= 503
+				if (lua_isinteger(L, index) != 0) {
+					return true;
+				}
+#endif
+				int isnum = 0;
+				const lua_Number v = lua_tonumberx(L, index, &isnum);
+				const bool success = isnum != 0 && static_cast<lua_Number>(std::llround(v)) == v;
 				if (!success) {
 					// expected type, actual type
 					handler(L, index, type::number, type_of(L, index));
@@ -6347,18 +6355,15 @@ namespace sol {
 		};
 
 		template<typename T>
-		struct getter<T, std::enable_if_t<meta::all<std::is_integral<T>, std::is_signed<T>>::value>> {
+		struct getter<T, std::enable_if_t<std::is_integral<T>::value>> {
 			static T get(lua_State* L, int index, record& tracking) {
 				tracking.use(1);
-				return static_cast<T>(lua_tointeger(L, index));
-			}
-		};
-
-		template<typename T>
-		struct getter<T, std::enable_if_t<meta::all<std::is_integral<T>, std::is_unsigned<T>>::value>> {
-			static T get(lua_State* L, int index, record& tracking) {
-				tracking.use(1);
-				return static_cast<T>(lua_tointeger(L, index));
+#if SOL_LUA_VERSION >= 503
+				if (lua_isinteger(L, index) != 0) {
+					return static_cast<T>(lua_tointeger(L, index));
+				}
+#endif
+				return static_cast<T>(std::llround(lua_tonumber(L, index)));
 			}
 		};
 
@@ -7107,16 +7112,25 @@ namespace sol {
 		struct check_getter<T, std::enable_if_t<std::is_integral<T>::value && lua_type_of<T>::value == type::number>> {
 			template <typename Handler>
 			static optional<T> get(lua_State* L, int index, Handler&& handler, record& tracking) {
-				int isnum = 0;
-				lua_Integer value = lua_tointegerx(L, index, &isnum);
-				if (isnum == 0) {
-					type t = type_of(L, index);
-					tracking.use(static_cast<int>(t != type::none));
-					handler(L, index, type::number, t);
-					return nullopt;
+#if SOL_LUA_VERSION >= 503
+				if (lua_isinteger(L, index) != 0) {
+					tracking.use(1);
+					return static_cast<T>(lua_tointeger(L, index));
 				}
-				tracking.use(1);
-				return static_cast<T>(value);
+#endif
+				int isnum = 0;
+				const lua_Number value = lua_tonumberx(L, index, &isnum);
+				if (isnum != 0) {
+					const auto integer_value = std::llround(value);
+					if (static_cast<lua_Number>(integer_value) == value) {
+						tracking.use(1);
+						return static_cast<T>(integer_value);
+					}
+				}
+				const type t = type_of(L, index);
+				tracking.use(static_cast<int>(t != type::none));
+				handler(L, index, type::number, t);
+				return nullopt;
 			}
 		};
 
@@ -7317,6 +7331,8 @@ namespace sol {
 
 // end of sol/raii.hpp
 
+#include <cassert>
+#include <limits>
 #ifdef SOL_CODECVT_SUPPORT
 #endif
 #ifdef SOL_CXX17_FEATURES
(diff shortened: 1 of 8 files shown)


Nava2:
Shouldn't runtime checks be gated behind `SOL_CHECK_ARGUMENTS`.

mrgreywater:
`SOL_CHECK_NUMBER_PRECISION` checks are currently opt-in, but I guess I could define it with `SOL_CHECK_ARGUMENTS` and make have another makro to disable it...

ThePhD:
SOL_CHECK_ARGUMENTS is the top-level safety macro that turns everything else on.

Nava2:
The build issue is related to a typo in `master`, btw.

mrgreywater:
One question that came up: When passing -1 from lua to sol, should it validate as std::uint32_t and then be automatically be converted to 0xFFFFFFFF? Right now I implemented it so if you pass negative numbers, they will not match with unsigned integers.

ThePhD:
@Nava2 Yeah, I'm still fixing strange bugs related to `container[-1]` not always throwing for some reason...9

Nava2:
> @Nava2 Yeah, I'm still fixing strange bugs related to container[-1] not always throwing for some reason...

Maybe as @mrgreywater mentioned as an aside, could it be a natural cast to 0xFFFFFFFF as a 32-bit `size_t`?

ThePhD:
I've been using `std::ptrdiff_t` specifically to avoid that issue, actually! But I might still have to vet a few places. Note that it also happens with `container[0]` (which is also wrong: lua indexes from 1, and even though I do `key < 1` to check for 0, it still doesn't seem to error proper...!)

mrgreywater:
I added some more test cases to show the current expected behaviour in the unsigned vs signed integers.
@ThePhD would be nice if you could fix the build error on the develop branch. Btw, test_state.cpp fails sometimes on linux:i386 and windows:x86 with luajit51. Not consistently, but sometimes over multiple runs.

mrgreywater:
rewrote the PR, because lua_pushinteger/lua_tointeger etc is also 'broken' for x64 with values > 53bit, e.g. it doesn't give us a possibility to check for overflow.

single/sol/sol.hpp
```diff
@@ -6899,16 +6906,24 @@ namespace sol {
...
-				if (isnum == 0) {
-					type t = type_of(L, index);
-					tracking.use(static_cast<int>(t != type::none));
-					handler(L, index, type::number, t);
-					return nullopt;
+#if SOL_LUA_VERSION >= 503
+				if (lua_isinteger(L, index) == 1) {
+					tracking.use(1);
+					return static_cast<T>(lua_tointeger(L, index));
+				}
```
ThePhD: lua_isinteger followed by lua_tointeger is a slight inefficiency: lua_tointegerx already checks if the number is an integer and performs the cast and `lua_tointeger` re-does this work in 5.3 since it's just a macro around `lua_tointegerx`.

single/sol/sol.hpp
```diff
@@ -6899,16 +6906,24 @@ namespace sol {
...
+					return static_cast<T>(lua_tointeger(L, index));
+				} 
+#endif
+				if (type_of(L, index) == type::number) {
+					const lua_Number value = lua_tonumber(L, index);
+					const auto integer_value = std::llround(value);
+					if (static_cast<lua_Number>(integer_value) == value) {
+						tracking.use(1);
+						return static_cast<T>(integer_value);
+					}
```
ThePhD: Same comment here: use `lua_tonumberx` and then check the `isnum` property, before performing the appropriate round-tripping checks.

single/sol/sol.hpp
mrgreywater: lua_tointegerx doesn't tell me if the number was an integer though...

single/sol/sol.hpp
mrgreywater: lua_tonumberx says true if the value on the lua stack is a string that is convertible to a number, which means it has to parse the string. lua_type will just let me know if the value on the stack is a number outright.

mrgreywater: I can change it, but I doubt it's faster or makes sense tbh.

mrgreywater:
Your notes open up the question: should sol accept strings as numbers if they are convertible to integers/floating points?

ThePhD:
When someone has `SOL_CHECK_ARGUMENTS` defined, the `stack::check` aggressively enforces the `number == number` requirement for nearly all operations. Without it, we have to be lenient: many in-the-wild codebases have this expectation of sol2.

As a side note not related to this PR, I'm thinking that since the request is popular enough I might add an "opt out" macro that allows someone to ignore the type safety rammifications of implicit `string` -> `number` concerns and for sol2 to not check them, provided they understand just what they're getting into (for example, no ability to overload on `string` versus `number` and essentially getting a first-come-first-serve style of overloading between those types).

ThePhD:
I don't test x86 because many of the tests throw and LuaJIT on x86 is a nightmare. LuaJIT 2.1.x-beta+ might have fixed it, but that doesn't really help me for older tests.

I have been slowly converting throwing tests into non-throwing variants, but that's also going to take a while. Assuring that it works or 64-bit and running the important tests locally on x86 will be enough for me for now while I try to get the test harness all beefed out.

mrgreywater:
Well, I think having a 32bit build would still be nice to have, maybe we can just have it so it's allowed to fail for now? It's kinda needed, since the lua_Integer has a different size there.
As for your review comments, can you let me know what handling you want for this PR? Should I allow strings as number or not? For either case I would need to add test cases and make sure it works in a well defined way across versions.

ThePhD:
It took literally all day, but I managed to build on windows:x86 with luajit 2.0.5 and purge most uses of exceptions being transported through the Lua C API. This mean intermittent and non-understandable failures due to SIGSEGV and other hard crashes should almost never happen anymore.

Woo

I also solved thread problems with CATCH (`REQUIRE` is not thread safe: none of CATCH is, actually...!), so that will solve the inconsistent `utility/thread` test. The other inconsistent tests I can't make repro anymore after the changes, so I can't really say anything about it.

I still need to look at `SOL_SAFE_EXCEPTIONS_PROPOGATION` under LuaJIT 2.1.3beta3 and see if I can safely default the macro to being on. Right now, however, I'm going to work on getting SOL_USING_CXX_LUA to not choke on `catch ( ... )` and hiding error messages from `luaL_error` tosses.

mrgreywater:
@ThePhD Nice. Please consider the [/EH](https://msdn.microsoft.com/en-us/library/1deeycx5.aspx) options, msvc might otherwise not expect the lua c functions to work with exceptions, and won't compile proper exception handlers.
Also what I noticed for older luajit versions together with msvc is that any function using lua_error may not throw (noexcept). If such a function exists, it must be split in two, with one being a c++ trampoline function before calling lua_error. Otherwise it will crash for sure if an error condition arises.

ThePhD:
I already said it earlier:

> When someone has SOL_CHECK_ARGUMENTS defined, the stack::check aggressively enforces the number == number requirement for nearly all operations. Without it, we have to be lenient: many in-the-wild codebases have this expectation of sol2.

The only place you should be doing hard `type == number` checks is `stack::check`, in other words. In every other functions, you must be lenient (e.g., in `check_getter`, gate the `type == number` check by `SOL_CHECK_ARGUMENTS`). This is because many libraries depend on `string` => `number` conversions and when porting to sol2 cannot afford to have this behavior reverted on them.

After we're done with this PR I'll worry about figuring out where to put a `SOL_LUA_NUMBERS_FROM_STRINGS` define.

mrgreywater:
What I meant by consider it is that if you want to officially support luajit 2.0, you might have to test different \EH flags, and depending on the outcome add documentation on requirements when compiling sol2+luajit2.0+msvc. sol2 might need to be configured differently when you use different \EH flags, or might not even work at all with exceptions.

mrgreywater:
Changed the type checks to lua_tonumberx where applicable, please let me know if you're happy with the changes or there's anything else to do.

ThePhD:
All tests seem to pass, I'll incorporate this PR and add any opt-in/optout macros myself.

Thanks for working on this so diligently, and sorry for having to let you rebase to upstream so many times. Trying to fix sol2 right now is a bit of a grudge match since I have so many things to overhaul, but for 99% of applications it's still okay even if the tests won't shut up.

iaz3:
Note: If anyone happens to include Windows.h before this, without defining NOMINMAX, it fails to compile, due to [these lines](https://github.com/ThePhD/sol2/commit/372db6f6abd4dde640a07996e69e2bc727d69cd4#diff-cedd1ed139bc62fa049e6365070939f4R7503) (possibly more? first one it choked on, at least)

A solution can be to either advise users to somehow remove the defintions, possibly using [push_macro](https://msdn.microsoft.com/en-us/library/2bs04eza.aspx?f=255&MSPPError=-2147217396) and [pop_macro](https://msdn.microsoft.com/en-us/library/hsttss76.aspx) to undefine min and max like so, if they need windows versions, or to do it ourselves

```C++
#pragma push_macro("min")
#pragma push_macro("max")  
#undef min
#undef max
// sol.hpp here ...
#pragma pop_macro("max")  
#pragma pop_macro("min")
```

I hate windows.h

ThePhD:
I wrapped every instance of `max` and `min` with parenthesis to prevent this probl- oh.

That's what I forgot to do with this PR...

PR #486, opened 2017-08-17
Renamed loop index to avoid hiding function parameter (simple_usertype_metatable.hpp)

eliasdaler:
Renamed loop index to j to avoid hiding function parameter


ThePhD:
@eliasdaler Ooh, thanks. Sorry about that: I didn't really get this warning (because building on W4 and Wall is noisy on Windows).

eliasdaler:
W4 gets okay if you disable this warnings in VS: 4503;4458;4127;4702

PR #493, opened 2017-08-24
[WIP] Add a CMakeLists.txt

alkino:
Make comments, test.
I like if we can make the all in one header with cmake.


Nava2:
This really needs to use some loops and separate lists. This is a great first attempt but the nightmare of maintenance this style creates is problematic. 

Since all the examples and tests have the same compiler options, you can create a LIST and iterate over the list to define all of the targets instead of doing each on its own line.

mrgreywater:
Not only loops, but you can use some general file(GLOB... to create the lists. Also consider source_group so the generated projects are properly organized for msvc. Additionally the compiler flags should match the settings of the current Ninja build. Finally it should probably find and link to whatever Lua version you choose via cmake options.

Nava2:
@alkino check out the work I started about a year ago here: 1a381ae4a1e5bbf0305028fbe76730d5114303c9.

It's quite a bit more organized, I don't mean to deter you from contributing. :)

alkino:
I'm not an expert of cmake. I did it only because @ThePhD thinks it was useful. Don't be shy to promote your work or modify my patch.

Nava2:
That is *much* better! I'll look a bit more, now.

CMakeLists.txt
```diff
@@ -6,152 +6,46 @@ set(CMAKE_CXX_STANDARD 14)
...
-        overloading
-        usertype_automatic_operators
-        containers_as_table
-        usertype_special_functions
-        stack_aligned_function
-        optional_with_iteration
-        config
-        functions
-        RUNTIME DESTINATION bin/examples)
+set(LUA_VERSION "5.3" CACHE STRING "The version of Lua needed (5.1, 5.2, 5.3)")
```
Nava2: Do you need to handle Lua JIT? IIRC they are found the same way, but just unsure if more checks are needed.

CMakeLists.txt
```diff
@@ -6,152 +6,46 @@ set(CMAKE_CXX_STANDARD 14)
...
-		test_simple_usertypes.cpp
-		test_gc.cpp
-		test_functions.cpp
-		test_usertypes.cpp
-         	test_containers.cpp
-        	test_filters.cpp)
-install(TARGETS tests RUNTIME DESTINATION bin)
+include_directories(${LUA_INCLUDE_DIRS})
+
+file(GLOB EXAMPLES_SRC examples/*.cpp)
```
Nava2: In my opinion, I prefer explicit source lists due to many tools being unable to support the globbed files. 

You can use the same machinery you use on L22 on with a List. This one is a matter of opinion, so someone else should likely weigh in.

CMakeLists.txt
alkino: Maybe we need to take your code here.

mrgreywater:
Can we build against the seperate sol source files, and run 'single.py' in the cmake build? Otherwise editing sol with tools like clion won't work properly.
@Nava2 which tools actually don't support globbing of cmake files? without globbing, it's just a hassle whenever you add or remove any files to the project

CMakeLists.txt
```diff
@@ -11,13 +11,58 @@ if (LUA_VERSION MATCHES "5.1")
...
   Find_Package(Lua 5.3 EXACT REQUIRED)
+else()
+  message(FATAL_ERROR "${LUA_VERSION} is not a supported version for lua.")
 endif()
 
 include_directories(${LUA_INCLUDE_DIRS})
 
-file(GLOB EXAMPLES_SRC examples/*.cpp)
+if (NOT MSVC)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-warning -Wno-unknown-warning-option -Wall -Wextra -Wpedantic -pedantic -pedantic-errors -Wno-noexcept-type -ftemplate-depth=1024")
```
mrgreywater: Is this tested for both gcc and clang? Maybe seperate release and debug flags?

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+             test_customizations.cpp test_large_integer.cpp
+             test_inheritance.cpp tests.cpp
+             test_variadics.cpp test_coroutines.cpp
+             test_container_semantics.cpp test_storage.cpp
+             test_overflow.cpp test_plain_types.cpp
+             test_simple_usertypes.cpp test_gc.cpp
+             test_functions.cpp test_usertypes.cpp
+             test_containers.cpp test_filters.cpp
+)
+source_group(Tests FILES ${TEST_SRC})
```
mrgreywater: should probably be lowercase 'tests' like 'examples'

CMakeLists.txt
ThePhD: gcc and clang both need all of these warning flags: there's 2 versions of unknown warning specifically because clang and gcc don't agree with each other. XCode uses a variant of appleclang but the flags remain the same.

CMakeLists.txt
Nava2: I agree with consistent, but I'd prefer all uppercase.

Nava2:
@mrgreywater, tools that don't use the CMake daemon can't have it work. Clion is nice enough to use it. KDevelop and QtCreator couldn't last I checked.

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+  Find_Package(Lua 5.1 EXACT REQUIRED)
+elseif(LUA_VERSION MATCHES "5.2")
+  Find_Package(Lua 5.2 EXACT REQUIRED)
+elseif(LUA_VERSION MATCHES "5.3")
+  Find_Package(Lua 5.3 EXACT REQUIRED)
+else()
+  message(FATAL_ERROR "${LUA_VERSION} is not a supported version for lua.")
+endif()
+
+include_directories(${LUA_INCLUDE_DIRS})
```
mrgreywater: should be `include_directories(${LUA_INCLUDE_DIR})`

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+
+project(sol2)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+include_directories(.)
+
+set(LUA_VERSION "5.3" CACHE STRING "The version of Lua needed (5.1, 5.2, 5.3)")
+if (LUA_VERSION MATCHES "5.1")
+  Find_Package(Lua 5.1 EXACT REQUIRED)
```
mrgreywater: should be `find_package` for consistensy

mrgreywater:
@Nada2 I tested both Kdevelop and QtCreator and they both worked fine with `file(GLOB`. So unless there is some other important IDE that for some reason is incompatible, I don't see why this should feature should be disregarded.

```
file(GLOB EXAMPLES_SRC ./examples/*.cpp)
file(GLOB TEST_SRC ./test_*.cpp ./test_*.hpp ./tests.cpp)
```

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+                 examples/usertype_special_functions.cpp
+                 examples/usertype_var.cpp
+                 examples/variables.cpp
+                 examples/variadic_args.cpp)
+source_group(examples FILES ${EXAMPLES_SRC})
+foreach(example_source_file ${EXAMPLES_SRC})
+message(STATUS "${example_source_file}")
+  get_filename_component(example_name ${example_source_file} NAME_WE)
+  add_executable(${example_name} ${example_source_file})
+  target_link_libraries(${example_name} ${LUA_LIBRARY})
```
mrgreywater: should be target_link_libraries(${example_name} ${LUA_LIBRARIES})

CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+             test_functions.cpp test_usertypes.cpp
+             test_containers.cpp test_filters.cpp
+)
+source_group(Tests FILES ${TEST_SRC})
+
+add_executable(tests ${TEST_SRC})
+target_include_directories(tests PRIVATE ./Catch/include/)
+
+find_package(Threads)
+target_link_libraries(tests Threads::Threads ${LUA_LIBRARY})
```
mrgreywater: LUA_LIBRARIES

CMakeLists.txt
mrgreywater: I think lowercase makes more sense, as the examples folder is lowercase aswell. But it doesn't really matter.

CMakeLists.txt
```diff
@@ -0,0 +1,183 @@
...
+                sol/userdata.hpp
+                sol/usertype_core.hpp
+                sol/usertype.hpp
+                sol/usertype_metatable.hpp
+                sol/usertype_traits.hpp
+                sol/variadic_args.hpp
+                sol/variadic_results.hpp
+                sol/wrapper.hpp)
+
+find_package(PythonInterp 3 REQUIRED)
```
mrgreywater: Can we make this optional? People might not have python, especially on windows and still should be able to compile everything imo.

alkino:
First, thanks mrgreywater for all this comments. I fix some of them.

For LuaJIT, there is no findLuajit. I add https://gitlab.kitware.com/cmake/cmake/issues/17204
Maybe we could do this ourself.

About globbing I don't really care. What is the consensus?

About single.py, I add it but it's not "automatic". I need to do "make single_sol && make install". Do you have any idea to generate it when doing "make install" or "make" alone.

About IDE and OS, I use only linux with vim. I have no access to other.

CMakeLists.txt
alkino: Yes we can.

CMakeLists.txt
```diff
@@ -0,0 +1,185 @@
...
+                sol/variadic_args.hpp
+                sol/variadic_results.hpp
+                sol/wrapper.hpp)
+
+find_package(PythonInterp 3)
+if (PYTHONINTERP_FOUND)
+  add_custom_command(OUTPUT sol.hpp COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/single.py --output ${CMAKE_CURRENT_BINARY_DIR}/sol.hpp DEPENDS ${HEADER_SRCS})
+  add_custom_target(single_sol DEPENDS sol.hpp)
+  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sol.hpp" DESTINATION include)
+endif()
```
mrgreywater: perhaps there could be a message if single sol was generated or not?

CMakeLists.txt
```diff
@@ -0,0 +1,185 @@
...
+source_group(TESTS FILES ${TEST_SRC})
+
+add_executable(tests ${TEST_SRC})
+target_include_directories(tests PRIVATE ./Catch/include/)
+
+find_package(Threads)
+target_link_libraries(tests Threads::Threads ${LUA_LIBRARIES})
+install(TARGETS tests RUNTIME DESTINATION bin)
+
+set(HEADER_SRCS sol/as_args.hpp
```
mrgreywater: you need to add them to the executable as well, like TEST_SRC or EXAMPLES_SRC

CMakeLists.txt
```diff
@@ -0,0 +1,185 @@
...
+                sol/usertype.hpp
+                sol/usertype_metatable.hpp
+                sol/usertype_traits.hpp
+                sol/variadic_args.hpp
+                sol/variadic_results.hpp
+                sol/wrapper.hpp)
+
+find_package(PythonInterp 3)
+if (PYTHONINTERP_FOUND)
+  add_custom_command(OUTPUT sol.hpp COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/single.py --output ${CMAKE_CURRENT_BINARY_DIR}/sol.hpp DEPENDS ${HEADER_SRCS})
```
mrgreywater: I think this command will replace the wrong sol.h it should be the one in the single folder

CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+             test_strings.cpp test_environments.cpp
+             test_customizations.cpp test_large_integer.cpp
+             test_inheritance.cpp tests.cpp
+             test_variadics.cpp test_coroutines.cpp
+             test_container_semantics.cpp test_storage.cpp
+             test_overflow.cpp test_plain_types.cpp
+             test_simple_usertypes.cpp test_gc.cpp
+             test_functions.cpp test_usertypes.cpp
+             test_containers.cpp test_filters.cpp
+)
```
mrgreywater: nitpicking here, but the closing bracket should be on the previous line if you want to be consistent :P

mrgreywater:
@alkino There are cmake modules like these: https://github.com/cactus/cronkite/blob/master/cmake/modules/FindLuaJIT.cmake which can find LuaJIT. Imo it would make sense to use something like that.
For the globbing, let's wait for @ThePhD @Nava2 to comment on that again.
With the single/sol/sol.hpp automatic command I don't really know. I would have to fiddle with that myself to try if it's possible.
I can test on different OS's (except Mac). I can try on all platforms before merging, it's just that cmake often needs some minor tweaking for each platform/compiler. e.g find_package probably won't really work on windows, so we need some other way to manually supply the desired lua path.

CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+add_executable(tests ${TEST_SRC} ${HEADER_SRCS})
+target_include_directories(tests PRIVATE ./Catch/include/)
+
+find_package(Threads)
+target_link_libraries(tests Threads::Threads ${LUA_LIBRARIES})
+install(TARGETS tests RUNTIME DESTINATION bin)
+
+find_package(PythonInterp 3)
+if (PYTHONINTERP_FOUND)
+  add_custom_command(OUTPUT single/sol.hpp COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/single && ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/single.py --output ${CMAKE_CURRENT_BINARY_DIR}/single/sol.hpp DEPENDS ${HEADER_SRCS})
```
mrgreywater: shouldn't it be ${CMAKE_CURRENT_SOURCE_DIR}/single/sol.hpp? Some IDE's use a custom build directory, so this will probably break like it is right now.

alkino: I'm not trying to put it in source directory. I write it in the build directory.
I think that you should never let build tools modify your sources.

mrgreywater: well, single/sol/sol.hpp is autogenerated, and just for user's conveniance. I don't see the point of having the single sol file automatically generated with cmake if you need to manually copy it over afterwards anyway. Then why don't just execute the command yourself in the first place.

alkino: hum.
The right way to me seems to make
`
mkdir BUILD && cd BUILD && cmake .. && make install DESTDIR=foo
`
and get the sol.hpp from here.
But it may be convenient to get the single/sol/sol.hpp without compiling it. (windows doesn't have python 3 easily as you said).
So I don't know. But generate source by default in source directory seems me awkward.

mrgreywater: well, if you compile it with the command line, sure. But consider the fact that you won't compile with the command line on windows, and also on linux alot of people use QtCreator, KDevelop or CLion. The whole point of having it generated in the cmake file is ease of use, having to copy it over afterwards defeats the whole purpose imo.

mrgreywater: Thinking about it more, maybe you could make it a cmake option. So people who really don't want it auto generated can opt-out of it.

Nava2:
I have a FindLuaJIT and FindLua module that works cross-platform. I'll donate those to the project. If globbing works now, I see no concern @mrgreywater!

CMakeLists.txt
```diff
@@ -0,0 +1,188 @@
...
+                 examples/usertype_simple.cpp
+                 examples/usertype_special_functions.cpp
+                 examples/usertype_var.cpp
+                 examples/variables.cpp
+                 examples/variadic_args.cpp)
+source_group(EXAMPLES FILES ${EXAMPLES_SRC})
+foreach(example_source_file ${EXAMPLES_SRC})
+message(STATUS "${example_source_file}")
+  get_filename_component(example_name ${example_source_file} NAME_WE)
+  add_executable(${example_name} ${example_source_file} ${HEADER_SRCS})
```
mrgreywater: can we have it so the examples are called like "example-variadic_args", "example-usertype" etc? Otherwise it gets difficult to find the test executable in the alphabetically ordered build target drop down lists of the ide's.

CMakeLists.txt
```diff
@@ -172,10 +171,10 @@ set(EXAMPLES_SRC examples/any_return.cpp
                  examples/usertype_var.cpp
                  examples/variables.cpp
                  examples/variadic_args.cpp)
-source_group(EXAMPLES FILES ${EXAMPLES_SRC})
+source_group(examples FILES ${EXAMPLES_SRC})
 foreach(example_source_file ${EXAMPLES_SRC})
-message(STATUS "${example_source_file}")
   get_filename_component(example_name ${example_source_file} NAME_WE)
+  set(example_name "example-${example_name}")
```
Nava2: Unless you use `example_name` again, there's no point using a variable. Just inline this into L178.

alkino: I use on several lines here. Not sure to understand.

Nava2: My mistake, ignore me. 😨

CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+                  cxx_range_for
+                  cxx_return_type_deduction
+                  cxx_right_angle_brackets
+                  cxx_static_assert
+                  cxx_strong_enums
+                  cxx_variadic_macros
+                  cxx_variadic_templates)
+
+file(GLOB HEADER_SRCS sol/*.hpp)
+set(HEADER_SRCS sol.hpp ${HEADER_SRCS})
```
mrgreywater: source_group?

CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
...
+
+if (NOT MSVC)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-warning -Wno-unknown-warning-option -Wall -Wextra -Wpedantic -pedantic -pedantic-errors -Wno-noexcept-type -ftemplate-depth=1024")
```
mrgreywater: We should have Release/Debug optimization flags. See my previous comment.

CMakeLists.txt
alkino: ok

CMakeLists.txt
alkino: The default target are not good enough?

`cmake . -DCMAKE_BUILD_TYPE=Release|Debug`

mrgreywater: CMake won't do anything on its own with the CMAKE_BUILD_TYPE. You still have to add the optimization flags yourself in the CMakeLists.txt. If you scroll up in the comments there is an example how it could look like.

alkino: https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/Compiler/GNU.cmake#L39
It does for me.

mrgreywater: Looks like I'm wrong, proceed :D

alkino:
So, if we disable tests and examples, single.py is not run. And we install an old sol.hpp.

And do you want sol.hpp installed on system in "include/sol/sol.hpp" or "include/sol.hpp"?

CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
+cmake_minimum_required(VERSION 3.0)
+
+project(sol2)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+include_directories(.)
```
dmerejkowsky: I prefer using `target_include_directories`

CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+if (EXAMPLES)
+  file(GLOB EXAMPLES_SRC examples/*.cpp)
+  source_group(examples FILES ${EXAMPLES_SRC})
+  foreach(example_source_file ${EXAMPLES_SRC})
+    get_filename_component(example_name ${example_source_file} NAME_WE)
+    set(example_name "example-${example_name}")
+    add_executable(${example_name} ${example_source_file} ${HEADER_SRCS})
+    target_link_libraries(${example_name} ${LUA_LIBRARIES})
+    target_compile_features(${example_name} PUBLIC ${CXX_FEATURES})
+    install(TARGETS ${example_name} RUNTIME DESTINATION bin/examples)
```
dmerejkowsky: you really want to distribute the 'built' examples ? wouldn't it be better to distribute the *sources* instead ?

CMakeLists.txt
alkino: As it is for all target I prefer a global one.
You talk me about https://cmake.org/cmake/help/latest/variable/CMAKE_INCLUDE_CURRENT_DIR.html I will try to dig it.

CMakeLists.txt
alkino: You're right.

CMakeLists.txt
Nava2: Even though its for all, there's an argument surrounding the idea of using the single header to build, too. If we have the single header it's a good test to use it to verify it works the same as the full tree.

CMakeLists.txt
Nava2: I'd just remove the install here.

Nava2:
For the install target, only enable it if python 3 is found and the single header can be generated. 

Additionally, I'd not install sources into include/ but into /usr/src (/src on windows). See http://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.html#USRSRCSOURCECODE

CMakeLists.txt
alkino: But the single one can only be generated if python3 is found.

Nava2: @ThePhD expressed in gitter he would like to see building implemented with the single header, too. Thus, if it's generated, we should be checking that they build with both the single and full tree. 

If this is running in CI, ie Travis or appveyor, we should make configuration fail if it can't find python.

Nava2:
I think we should close this PR and start a new. It's gotten rather convoluted now and we are getting to a point that its not a WIP anymore, but nearing merge quality!

mrgreywater:
He can just squash everything, no need to create a new PR oO

mrgreywater:
tests are now in a subfolder. 
@alkino can you fix the remaining issues and maybe squash the commits?

alkino:
Done.
I think we should squash and merge it. (it's possible with github). Be careful to change the name of the commit and not of the ticket.

ThePhD:
I haven't forgot about this, I've just been trying to manage everything else!

Thanks for the hard work on this. Once I integrate it I'll finally be able to set up all the tests.

Nava2:
I have changes to how the Lua configs work from the upstream requests. I'll add those in later, though. LGTM, assuming @mrgreywater agrees.

CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+
+project(sol2)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+include_directories(.)
+
+set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules" "${CMAKE_MODULE_PATH}")
+
+set(LUA_VERSION "5.3" CACHE STRING "The version of Lua needed (5.1, 5.2, 5.3, LuaJIT)")
```
Nava2: Change `LUA_VERSION` -> `SOL_LUA_VERSION`. Don't mess with the names created by CMake itself.

CMakeLists.txt
Nava2: Remove this install.

CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+    install(TARGETS ${example_name} RUNTIME DESTINATION bin/examples)
+  endforeach()
+endif()
+
+if (TESTS)
+  file(GLOB TEST_SRC tests/test*.cpp)
+  source_group(tests FILES ${TEST_SRC})
+
+  add_executable(tests ${TEST_SRC} ${HEADER_SRCS})
+  target_include_directories(tests PRIVATE ./Catch/include/)
```
Nava2: We should replace the in-tree catch with [`ExternalProject_Add(...)`](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html): 

```cmake
ExternalProject_Add(
    catch
    PREFIX ${SOL_VENDOR_PREFIX}
            
    GIT_REPOSITORY "https://github.com/philsquared/Catch.git"
    GIT_TAG tags/v1.3.4 # Update version here
    
    CONFIGURE_COMMAND ""
    BUILD_COMMAND ""
    INSTALL_COMMAND ""
    TEST_COMMAND "")
```

CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+
+  find_package(Threads)
+  target_link_libraries(tests Threads::Threads ${LUA_LIBRARIES})
+  install(TARGETS tests RUNTIME DESTINATION bin)
+endif()
+
+find_package(PythonInterp 3)
+if (PYTHONINTERP_FOUND)
```
Nava2: Add condition that if this is on `CI` (aka use `ENV{CI}`) we fail the build. Not using the single header isn't okay on CI.

CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+  add_custom_command(OUTPUT single/sol/sol.hpp COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/single.py --output ${CMAKE_CURRENT_SOURCE_DIR}/single/sol/sol.hpp DEPENDS ${HEADER_SRCS})
+  add_custom_target(single_sol DEPENDS single/sol/sol.hpp)
+  install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/single/sol/sol.hpp" DESTINATION include/sol)
+  message(STATUS "single_sol can be generated as python 3 has been found.")
+else()
+  message(STATUS "single_sol cannot be generated as python 3 has not been found.")
```
Nava2: `STATUS` -> `WARNING`

cmake/Modules/FindLua.cmake
```diff
@@ -0,0 +1,233 @@
...
+
+unset(_lua_include_subdirs)
+unset(_lua_append_versions)
+unset(_lua_library_names)
+
+
+include(${CMAKE_CURRENT_LIST_DIR}/FindLua/set_version_vars.cmake)
+include(${CMAKE_CURRENT_LIST_DIR}/FindLua/version_check.cmake)
+
+# # this is a function only to have all the variables inside go away automatically
```
Nava2: This is totally *my* fault, but remove these commented functions.

Nava2:
There still needs to be single vs. source-tree targets created. The examples should be built with the single header, the tests should be built against both (and tested).

PR #515, opened 2017-09-30
allocate userdata with proper alignment

### LINKED ISSUE #514: Alignment error with member callback
The following example, where we try to execute a `std::function` member object, creates alignment errors on clang 6.0.0 according to the UndefinedBehaviorSanitizer.

```
#include <sol.hpp>
#include <functional>
#include <iostream>

struct Test{
  std::function<void()> callback = [](){ std::cout << "Hello world!" << std::endl; };
};

int main() {
  sol::state lua;
  lua.new_usertype<Test>("Test","callback",&Test::callback);
  Test test;
  lua["test"] = &test;
  lua.script("test.callback()");
  return 0;
}
```

when compiled using `clang++ --std=c++14 -fsanitize=undefined ...` a cascade of 'misaligned address' runtime errors is generated, beginning with

``` 
(a.out:x86_64+0x1000d8c48): runtime error: constructor call on misaligned address 0x7fc01c405518 for type 'sol::function_detail::functor_function<std::__1::function<void ()> >', which requires 16 byte alignment
0x7fc01c405518: note: pointer points here
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
              ^ 
```  

Otherwise the program executes as expected.

mrgreywater:
This PR should fix #514.

Basically if we detect that the memory returned by lua_newuserdata isn't aligned after sol adds it bookkeeping, reallocate and align it manually. This should only happen rarely though. The only other alternatives to this approach are either 
- allocating seperately with another aligned allocator like `new`, but then we need a `__gc` metamethod for everything to be able call `delete` afterwards.
- or always allocating more space than we really need to be able to align the memory afterwards. (performance vs memory tradeoff)

Would love a review, it's easy to miss dependencies when tinkering with the inner workings of the library.

### DIFF
--- a/sol/stack_core.hpp
+++ b/sol/stack_core.hpp
@@ -51,6 +51,19 @@ namespace sol {
 		using unique_destructor = void (*)(void*);
 
 		template <typename T>
+		int user_aligned_alloc(lua_State* L, std::size_t offset, void*& mem, T*& obj) {
+			const auto alignment = std::alignment_of<T>();
+			mem = lua_newuserdata(L, offset + sizeof(T) + alignment - 1);
+			auto aligned_userdata = (std::uintptr_t)mem + offset;
+			// check the alignment. if it's not correct, adjust the pointer.
+			if ((aligned_userdata % alignment) != 0) {
+				aligned_userdata += alignment - aligned_userdata % alignment;
+			}
+			obj = reinterpret_cast<T*>(aligned_userdata);
+			return 1;
+		}
+
+		template <typename T>
 		inline int unique_destruct(lua_State* L) {
 			void* memory = lua_touserdata(L, 1);
 			T** pointerpointer = static_cast<T**>(memory);
@@ -62,9 +75,12 @@ namespace sol {
 		template <typename T>
 		inline int user_alloc_destruct(lua_State* L) {
 			void* rawdata = lua_touserdata(L, 1);
-			T* data = static_cast<T*>(rawdata);
+			auto data = (std::uintptr_t)rawdata;
+			if ((data % std::alignment_of<T>()) != 0) {
+				data += std::alignment_of<T>() - data % std::alignment_of<T>();
+			}
 			std::allocator<T> alloc;
-			alloc.destroy(data);
+			alloc.destroy((T*)data);
 			return 0;
 		}
 
--- a/sol/stack_push.hpp
+++ b/sol/stack_push.hpp
@@ -73,12 +73,12 @@ namespace stack {
 			// data in the first sizeof(T*) bytes, and then however many bytes it takes to
 			// do the actual object. Things that are std::ref or plain T* are stored as
 			// just the sizeof(T*), and nothing else.
-			T** pointerpointer = static_cast<T**>(lua_newuserdata(L, sizeof(T*) + sizeof(T)));
-			T*& referencereference = *pointerpointer;
-			T* allocationtarget = reinterpret_cast<T*>(pointerpointer + 1);
-			referencereference = allocationtarget;
+			T* obj = nullptr;
+			void* userdata = nullptr;
+			detail::user_aligned_alloc<T>(L, sizeof(T*), userdata, obj);
+			*reinterpret_cast<T**>(userdata) = obj;
 			std::allocator<T> alloc{};
-			alloc.construct(allocationtarget, std::forward<Args>(args)...);
+			alloc.construct(obj, std::forward<Args>(args)...);
 			f();
 			return 1;
 		}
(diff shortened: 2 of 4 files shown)


mrgreywater:
looks like libstd is missing std::align on quite a few targets, got to implement my own, i guess.

ThePhD:
You're calling that new detail  function for the deleters, which creates a new userdata, not just aligns the pointers. This behavior is leaking memory.

mrgreywater:
huh? I only replaced existing new_userdata calls...

tests/test_storage.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <catch.hpp>
 #include <sol.hpp>
+#include <cstdlib>
```
mrgreywater: can be removed

sol/stack_push.hpp
```diff
@@ -164,11 +164,16 @@ namespace stack {
 
 		template <typename... Args>
 		static int push_deep(lua_State* L, Args&&... args) {
-			P** pref = static_cast<P**>(lua_newuserdata(L, sizeof(P*) + sizeof(detail::unique_destructor) + sizeof(Real)));
+
+			Real* mem = nullptr;
+			void* userdata = nullptr;
+			detail::user_aligned_alloc<T>(L, sizeof(P*) + sizeof(detail::unique_destructor), userdata, mem);
```
mrgreywater: should be detail::user_aligned_alloc<Real>

sol/stack_core.hpp
```diff
@@ -51,6 +51,26 @@ namespace sol {
 		using unique_destructor = void (*)(void*);
 
 		template <typename T>
+		int user_aligned_alloc(lua_State* L, std::size_t offset, void*& mem, T*& obj) {
+			const auto alignment = std::alignment_of<T>();
+			std::size_t space = sizeof(T);
+			mem = lua_newuserdata(L, offset + space);
+			std::uintptr_t aligned_userdata = (std::uintptr_t)((char*)mem + offset);
```
mrgreywater: std::uintptr_t aligned_userdata = (std::uintptr_t)mem + offset;

ThePhD:
Oh sorry, you're right: I was a bit messed up due to github collapsing the lines between `unique_destruct` and the start of the changed code. There's no leak.

Still, if the allocators were changed, then the deallocators need to also be changed.

mrgreywater:
What deallocators do you mean? The userdata is deallocated automatically by the lua garbage collector.

TheLartians:
On my system the code from #514 now raises a `unaligned allocation.` exception. You probably still need to align the pointer in the 'dumb pusher' and the according getter method.

TheLartians:
> or always allocating more space than we really need to be able to align the memory afterwards.

I am strongly in favour of this as it leads to more predictable performance and the extra memory requirement is usually very low. Or at least let the user to decide via external macros.

mrgreywater:
@TheLartians 
Doing alignment in the dumb pusher will be tricky, but I'll check what I can do. If you would implement an aligned lua_Alloc function and pass it to the sol::state constructor, it would probably work though.

I'll change it to always allocate more memory for alignment purposes.

ThePhD:
lua_Alloc will not help you here. The problem will always be where the pointer data is, plus the data that comes after it. `lua_Alloc` only takes a size parameter to let you do the allocation book keeping: it doesn't dictate where the user of the C API will place their things in said memory.

mrgreywater:
@ThePhD not for the dumb pusher

mrgreywater:
ok, I removed the single/sol/sol.hpp, to avoid constant merge conflicts, changed it so we'll always allocate more space if the required alignment is >1.
Also the dumb pusher now allocates with the correct alignment, but it changes the memory layout of such userdata, and I'm not sure if there's any other place in sol that still requires the old memory layout. I quick look over all lua_touserdata didn't yield anything suspicious though.

mrgreywater:
@ThePhD Is there any particular reason for 
```cpp
			std::allocator<T> alloc;
			alloc.destroy((T*)data);
```
instead of
```cpp
((T*)data)~T();
```
And similarly `alloc.construct` vs `new (data)T()` ?

ThePhD:
Paranoia.

ThePhD:
This was implemented in https://github.com/ThePhD/sol2/commit/73484bf8e9eb6b8c36532e6e8f088a5f8f4e9566

The implementation also updates the docs, adds a test, and provides alignment even for misbehaved allocators.

PR #523, opened 2017-10-22
Fix syntax error in tutorial/variables example code.

aixxe:


ThePhD:
Peculiar. We compile the examples and run them in our tests: how did the compiler not choke on this...?

Thanks for catching.

PR #553, opened 2017-12-19
Use stack::push to push new_table in basic_table_core

a1ien:
Use stack::push to push new_table in basic_table_core instead call lua_createtable


ThePhD:
Out of curiosity, what problem does this fix? Or is this just for the fun of it?

a1ien:
It's just for fun. And for the consistency of the code

PR #565, opened 2018-01-22
Use std::bitset instead std::array of bool

a1ien:
For reduce memory footprint.
sizeof(std::array<bool, 32>) -> 32
sizeof(std::bitset<32>) -> 8 on x64, 4 on x32


ThePhD:
I never thought `std::bitset` was in C++03; I always thought it was a super recent addition! Well, let's go with it then 👍

Issue #541, opened 2017-11-29
sol::thread::status() returns incorrect value


THE-FYP:
Function `thread::status` always returning `thread_status::dead` if thread was runned.
```cpp
sol::state lua;
lua.open_libraries(sol::lib::coroutine);
lua.script("function main() coroutine.yield() end");
sol::thread t = sol::thread::create(lua);
sol::coroutine cor = t.state()["main"];
cor();
assert(cor.status() == sol::call_status::yielded); // ok
assert(t.status() == sol::thread_status::yielded); // error
```
Without [this condition](https://github.com/ThePhD/sol2/blob/e17455f5b67a96f35920772f21773d43f3f7aa66/sol/thread.hpp#L175) the returning thread status is correct.

Lua 5.3, LuaJIT 2.0.5, VS2017, latest sol2.

ThePhD:
Sounds about right. I'll take a look at the logic again. I'll probably just toss the condition out entirely or something.

ThePhD:
It's fixed in the latest.

PR #587, opened 2018-02-11
Fix thread::status() never returns a "dead" status

THE-FYP:
After fix of issue #541 a new bug has appeared - thread::status() returns `yielded` instead of `dead` when thread is dead. This PR fixes this problem in such a way as it's done in Lua: https://github.com/lua/lua/blob/master/lcorolib.c#L110

Test code:
```cpp
sol::state lua;
lua.open_libraries(sol::lib::coroutine);
lua.script("function main() print('start') coroutine.yield() print('end') end");
sol::thread t = sol::thread::create(lua);
sol::coroutine cor = t.state()["main"];
cor();
cor(); // thread should be dead after this call
assert(t.status() == sol::thread_status::dead); // but it isn't
cor(); // and here the coroutine starts over again instead of throwing "cannot resume dead coroutine"
```

### DIFF
--- a/sol/thread.hpp
+++ b/sol/thread.hpp
@@ -173,11 +173,15 @@ namespace sol {
 
 		thread_status status() const {
 			lua_State* lthread = thread_state();
-			thread_status lstat = static_cast<thread_status>(lua_status(lthread));
-			int stacksize = lua_gettop(lthread);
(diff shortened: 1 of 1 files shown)


ThePhD:
Thanks for the fix.

PR #591, opened 2018-02-17
Docs note that MSVC Compilers may require /bigobj

reworks:
So when making heavy use of templates when creating alot of usertypes, MSVC will error with ["Fatal Error C1128"](https://msdn.microsoft.com/en-us/library/8578y171.aspx). The solution is to use the [/bigobj](https://msdn.microsoft.com/en-us/library/ms173499.aspx) flag.


ThePhD:
... Why would this ever become a problem? Who would be using Visual C++ 2005 with sol2? It's not even possible to compile.

Why does this need to be apart of sol2's documentation?

reworks:
No, I meant that it causes errors on the latest versions of MSVC. I put that note there about 2005 just in case but it should probably be removed.

ThePhD:
Ah, alright. Well, it's been merged. Thanks for helping with the docs.

PR #628, opened 2018-04-13
 Compile fix for std::variant detection on Apple clang 9.1.0 (Xcode 9.3)

hzwar:
- Recent commit 80df3fc915a47087d8ae0831768b29c1171e6da2 didn't compile because of the __has_include check
 fixed according to https://clang.llvm.org/docs/LanguageExtensions.html#has-include
- Added SOL_STD_VARIANT ifdefs where variant is actually used

### DIFF
--- a/sol/stack_check.hpp
+++ b/sol/stack_check.hpp
@@ -31,9 +31,11 @@
 #include <functional>
 #include <utility>
 #include <cmath>
+#ifdef SOL_CXX17_FEATURES
 #ifdef SOL_STD_VARIANT
 #include <variant>
-#endif // C++17 variant
+#endif // SOL_STD_VARIANT
+#endif // SOL_CXX17_FEATURES
 
 namespace sol {
 namespace stack {
@@ -575,6 +577,7 @@ namespace stack {
 		}
 	};
 
+#ifdef SOL_CXX17_FEATURES
 #ifdef SOL_STD_VARIANT
 	template <typename... Tn, typename C>
 	struct checker<std::variant<Tn...>, type::poly, C> {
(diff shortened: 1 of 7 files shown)


sol/stack_check.hpp
```diff
@@ -31,9 +31,9 @@
 #include <functional>
 #include <utility>
 #include <cmath>
-#ifdef SOL_CXX17_FEATURES
+#ifdef SOL_STD_VARIANT
```
ThePhD: The SOL_CXX_17_FEATURES check should still be around this bllock: both should be present.

sol/stack_check.hpp
```diff
@@ -575,7 +575,7 @@ namespace stack {
 		}
 	};
 
-#ifdef SOL_CXX17_FEATURES
+#ifdef SOL_STD_VARIANT
```
ThePhD: Similar to above: both checks should be present.

sol/types.hpp
```diff
@@ -980,10 +982,10 @@ namespace sol {
 		template <>
 		struct lua_type_of<meta_function> : std::integral_constant<type, type::string> {};
 
-#ifdef SOL_CXX17_FEATURES
+#ifdef SOL_STD_VARIANT
```
ThePhD: Similarly here.

sol/stack_check_get.hpp
```diff
@@ -144,7 +144,7 @@ namespace stack {
 		}
 	};
 
-#ifdef SOL_CXX17_FEATURES
+#ifdef SOL_STD_VARIANT
 	template <typename... Tn>
```
ThePhD: Here as well: both `#ifdef` should be present.

ThePhD:
Thank you for your help!

hzwar:
I have to thank!

PR #636, opened 2018-04-25
Add support for plain arrays in sol::as_table

hmenke:
I hope my contribution meets the standards.
```cpp
#define SOL_CHECK_ARGUMENTS 1
#include <sol.hpp>

int main() {
    sol::state lua;

    double d[] = { 11, 22, 33 };
    lua["test"] = sol::as_table(d);
    lua.script("for i,v in ipairs(test) do print(i,v) end");
    // It just werks!
}
```

### DIFF
--- a/sol/stack_core.hpp
+++ b/sol/stack_core.hpp
@@ -529,6 +529,11 @@ namespace sol {
 				return static_cast<int>(c.size());
 			}
 
+			template <typename C, size_t N>
+			static int get_size_hint(const C (&c)[N]) {
+				return static_cast<int>(N);
+			}
+
 			template <typename V, typename Al>
 			static int get_size_hint(const std::forward_list<V, Al>&) {
 				// forward_list makes me sad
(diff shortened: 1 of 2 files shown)


ThePhD:
Would you mind adding some tests to the e.g. `test_tables` file or the `test_containers` file for this?

Also, I'm slightly concerned that it performs a copy by-default...

hmenke:
Actually the `as_table_t` type will always create a copy.  Say we have for example
```cpp
std::vector<double> d = { 11, 22, 33 };
lua["test"] = sol::as_table(d);
```
then the template parameter `T` will be deduced as `std::vector<double>`, i.e. `as_table_t`'s `source` member has a value type (rather than a reference).  That means that a copy will always occur.
https://github.com/ThePhD/sol2/blob/3620c4b8e1651a6b8bb39d2602e116551ee27026/sol/types.hpp#L385-L406

On the other hand I was wondering if this pull request is all that useful.  Actually it might be more useful to extend the `as_table` method to be able to push anything that supports `std::begin` and `std::end` (or rather any ADL free functions `begin` and `end`).  Initially I though it would be as easy as adapting the `has_begin_end` and `get_size_hint` traits but then I run into some ambiguity error.

PR #658, opened 2018-06-06
Add meson build configuration files

OrfeasZ:
This PR introduces meson build configuration files for sol2, allowing the library to be used as a dependency by other meson projects. When configuring with meson, the following two options are supported:

- `single` (default: `false`): If set to `true`, the single sol2 header will be generated and used instead of the multi-file configuration (requires python to be installed an available).
- `lua_cpp` (default: `false`): If set to `true` then lua will be compiled in C++ mode and `SOL_USING_CXX_LUA=1` will be defined.

Example of usage from another meson project (assumes sol2 is checked out in the subprojects directory):
```python
project('myproject', 'cpp')

sol2 = subproject('sol2', default_options: [ 'lua_cpp=true' ])
sol2_dep = sol2.get_variable('sol2_dep')

myproject = executable('myproject', 
    sources: [ 'myproject.cpp' ],
    dependencies: [ sol2_dep ],
)
```


ThePhD:
I am unsure if we should include the `lua.wrap` as part of this. sol2 shouldn't be providing a Lua to use: it should be given by the user. I don't know how Meson does this: is this wrap file a "default if there is nothing", or are we forcing lua 5.3.4 ?

OrfeasZ:
Currently the project is configured to use the system lua if it exists and fall back to using the wrapped lua if it doesn't. The only caveat is that when using the system lua we can't support C++ builds.

OrfeasZ:
I'll look into adding some tests.

ThePhD:
You don't have to add the meson tests to this PR: just open up a new one when you have the time @OrfeasZ

PR #662, opened 2018-06-11
Fix Exceptions Always Being Printed

### LINKED ISSUE #652: Bug report :SOL_PRINT_ERRORS  default print any exceptions / errors that occur in Release mode
In the newest sol2 sol.hpp release (2.20.0) there is a small bug in these code ,SOL_PRINT_ERRORS always defined,  neither 1 or 0.
```cpp
#ifdef SOL_PRINT_ERRORS
			std::cerr << "[sol2] An exception occurred: ";
			std::cerr.write(what.data(), what.size());
			std::cerr << std::endl;
#endif
```

ahoischen:
This PR fixes #652, which is about exceptions are always being printed. This was a very simple fix for a pretty inconsequential problem for most users. However, my application could really use those few KB freed by not needing `iostream`.

I'm not sure how I should update `single/`. I currently have it as a separate commit, but I can squash or drop it, to avoid clutter.

### DIFF
--- a/single/sol/sol.hpp
+++ b/single/sol/sol.hpp
@@ -20,8 +20,8 @@
 // CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 // This file was generated with a script.
-// Generated 2018-05-22 19:42:19.603781 UTC
-// This header was generated with sol v2.20.2 (revision d67c5b7)
+// Generated 2018-06-11 18:42:24.238190 UTC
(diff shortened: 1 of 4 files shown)


ThePhD:
I fixed this in my big update that I'm still working on, but honestly there's no reason to stay blocked. And yes, you update the single and include it in your PR: what I should do is auto-generate the header and then make my testing suite re-generate the header if all tests pass and then commit it into the repo...

But that's an idea for a fanciful, cool future I'm not gonna work on. Thanks for the help!

PR #675, opened 2018-06-22
fix sol2-config.cmake.in

Milerius:
It looks like there was a typo error, set a variable to SOL_INCLUDE_DIRS but use SOL2_INCLUDE_DIRS. And it is also possible that the installer of the library does not use sol2_single and if you do not activate the option SINGLE, you can not use it so you have to check that the target exists before.

This pull request will allow external c++ package manager to use sol2 properly with CMake.


Milerius:
see Microsoft/vcpkg#3766

PR #704, opened 2018-09-13
Fix unused variable warning

eliasdaler:
This fixes unused variable warning by putting "v" variable declaration under #ifdef.


eliasdaler:
Unfortunately, this PR introduced subtle bug, so I'll go with another approach to solving it in another PR.

PR #718, opened 2018-10-17
Pop stack element used as argument of lua_topointer

### LINKED ISSUE #717: sol::basic_reference::pointer leaves stack imbalanced
`sol::basic_reference::pointer` leaves the stack with one more element than was on it when called. `lua_topointer` does not pop from the stack, it merely returns the pointer to the top. So, after `sol::basic_reference::pointer` is called, there is reference to the object `pointer` was called on the stack. Simple test case:

```c++
#include <sol.hpp>
#include <cassert>
#include <stdio.h>

struct vars {
    int boop = 0;
};

int main() {
    sol::state lua;
    lua.new_usertype<vars>("vars", "boop", &vars::boop);
    lua["obj"] = boop();
    auto ref = lua["obj"];
    int stackSize = lua.stack_top();
    const void* ptr = ref.pointer();
    printf("obj: %p\n", ptr);
    assert(lua.stack_top() == stackSize);
}
```

jquesnelle:
This fixes #717 by popping the stack element that was pushed to be the argument of `lua_topointer`

### DIFF
--- a/sol/reference.hpp
+++ b/sol/reference.hpp
@@ -432,6 +432,7 @@ namespace sol {
 		const void* pointer() const noexcept {
 			int si = push();
 			const void* vp = lua_topointer(lua_state(), -si);
+			pop();
 			return vp;
 		}
 


jquesnelle:
Ah yeah, you're right. 2ebbe5d pops the right item

PR #724, opened 2018-10-28
Inside sol::proxy::call, check if we're building with clang

geneotech:
If building with clang on Windows, it tries to imitate MSVC by specifying reasonable _MSC_VER defines.
This makes sol2 compile MSVC version of sol::proxy::call (without the "template" preffix) that is nevertheless not accepted by clang.

This commit fixes the problem by also checking if "__clang__" is defined.


ThePhD:
There's one other place this check is at: make sure you put it in load_result,hpp, then regenerate the single header that's in the repo.

geneotech:
Done! Looks like many things changed after the regeneration, I hope I didn't it screw up, from the repository's I just run:

``./single.py --output single/sol/sol.hpp``

PR #770, opened 2019-02-13
latest require more parameters

RPG3D:


ThePhD:
Wow, Unreal again. Never expected to see this in my dash.

I think I'm going to finally file a bug report with Epic. It's a little silly I have to redefine their macro, and it's incredibly unhygienic of them to let it escape their headers.

ThePhD:
I solved this using what was suggested in https://github.com/ThePhD/sol2/issues/771 instead. Thanks, though!

PR #794, opened 2019-03-23
Don't use <lua.hpp>

Cannedfood:
The <lua.hpp> header is not included in some lua distributions, in particular www.github.com/lua/lua.

### DIFF
--- a/sol/compatibility/version.hpp
+++ b/sol/compatibility/version.hpp
@@ -1,4 +1,4 @@
-// sol2 
+// sol2
 
 // The MIT License (MIT)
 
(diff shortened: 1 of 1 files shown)


ThePhD:
Counterpoint: they should have `lua.hpp` and you should petition them to add it.

ThePhD:
... In fact, it was supposed to be part of that distribution. Did they remove it??

ThePhD:
Oh, it's part of the official distribution, just not that mirror. And to be quite honest, I'm not sure that repo is for anything other than visible history.

I'd rather not.

Cannedfood:
According [mailing list](http://lua-users.org/lists/lua-l/2004-09/msg00559.html) they won't add it. 
And it is somewhat annoying/ugly to work around it if you're using the git mirror.

ThePhD:
This has hooks for detection in sol3, but I'm not really interested in porting this to sol2: https://github.com/ThePhD/sol2/blob/sol3/include/sol/compatibility/version.hpp#L40

PR #854, opened 2019-08-07
No more LUA_ERRGCMM errors

sumneko:


ThePhD:
Just one comment: don't define this is `compat-5.3.h`, put this in a new file `compat-5.4.h`, since this is for lua 5.4. Then include that with the necessary checks in the header.

include/sol/compatibility/compat-5.3.h
```diff
@@ -418,4 +424,3 @@ COMPAT53_API void luaL_requiref(lua_State *L, const char *modname,
 
 
 #endif /* KEPLER_PROJECT_COMPAT53_H_ */
-
```
ThePhD: Please leave the blank lines at the end of the files there.

sumneko:
> Just one comment: don't define this is `compat-5.3.h`, put this in a new file `compat-5.4.h`, since this is for lua 5.4. Then include that with the necessary checks in the header.

Sorry, it's a little difficult for me, because I know little about C/C++ .
Maybe change this to issue ?

PR #866, opened 2019-09-13
Enable C++17 features when __cplusplus >= 201703.

dpacbach:
This will keep them enabled even when compiling with later
standards or experimental standards such as -std=c++2a.

Prior to this it was requiring strict equality.


ThePhD:
Thanks.

PR #898, opened 2019-11-13
Static analysis warnings

minirop:
Hello,
I've run some static analysis on my project and a few warnings from sol showed up.
So I fixed the obvious ones.

### DIFF
--- a/include/sol/stack_core.hpp
+++ b/include/sol/stack_core.hpp
@@ -1396,7 +1396,6 @@ namespace sol {
 				}
 				else {
 					if constexpr (std::is_same_v<std::equal_to<>, Op> // clang-format hack
-					     || std::is_same_v<std::less_equal<>, Op>     //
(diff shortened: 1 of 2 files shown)


ThePhD:
Thank you for taking a look into this?

Still, the static analyzer is doing something weird when tracing those template instantiations: that should not be `v.second` in the container stuff. The rest looks okay, though.

Can you rebase without the `.second` change, and then push those changes? That'll get the PR bots to try all the tests on your code again.

minirop:
the second commit was on my side. the analyzer was saying "variable set 2 times in a row" so looking at other methods I thought it was missing something. (should have run the tests, my bad)
Was it done on purpose or couldn't you do something like `*it = value.as<V>();`?

the full method for context:
```cpp
decltype(auto) v = *it;
v = value.as<V>();
return {};
```

ThePhD:
It was done on purpose.

Thanks for the rebase, it all looks clean now! 🎉

PR #925, opened 2020-01-28
Uniform management of optional types (sol::optional, std::optional)

Smertig:
Let's consider the following program:

```
#include <sol/sol.hpp>
#include <iostream>

template <typename T>
const char* f(T value) {
	return value.has_value() ? "fail" : "ok";
}

int main() {
	sol::state L;

	#define TEST(T) L.set_function("f", f<T>); std::cout << #T << ": " << static_cast<std::string>(L["f"]()) << "\n";

	TEST(sol::optional<int>);
	TEST(std::optional<int>);
	TEST(sol::optional<sol::table>);
	TEST(std::optional<sol::table>);
	TEST(sol::optional<sol::function>);
	TEST(std::optional<sol::function>);
	TEST(sol::optional<sol::object>);
	TEST(std::optional<sol::object>);

	return 0;
}
```

Current output is:

> sol::optional\<int\>: ok
> std::optional\<int\>: ok
> sol::optional\<sol::table\>: ok
> std::optional\<sol::table\>: ok
> sol::optional\<sol::function\>: ok
> std::optional\<sol::function\>: **fail**
> sol::optional\<sol::object\>: ok
> std::optional\<sol::object\>: **fail**

However, there should be no **fail** at all, because we call `f<T>` without an argument, so `value` must be `nullopt`, but it's not. It looks like `sol2` incorrectly handles some types wrapped in `std::optional`, although there is no problem with `sol::optional`.

This PR unifies management of both `sol::optional<T>` and `std::optional<T>` using `meta::is_optional` trait and thus fixes the previous example. I've also added some generic tests for both optionals.

The next step is to get rid of `meta::is_optional`. It can be replaced with some `is_detected`-based solution to completely remove dependency of `std::optional` and weird `#ifdef`s

### DIFF
--- a/include/sol/stack_check_get_qualified.hpp
+++ b/include/sol/stack_check_get_qualified.hpp
@@ -57,9 +57,10 @@ namespace sol { namespace stack {
 		}
 	};
 
-	template <typename T>
-	struct qualified_getter<optional<T>> {
-		static optional<T> get(lua_State* L, int index, record& tracking) {
+	template <typename O>
+	struct qualified_getter<O, std::enable_if_t<meta::is_optional_v<O>>> {
+		static O get(lua_State* L, int index, record& tracking) {
+			using T = typename O::value_type;
 			if constexpr (is_lua_reference_v<T>) {
 				// actually check if it's none here, otherwise
 				// we'll have a none object inside an optional!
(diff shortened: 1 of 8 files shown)


ThePhD:
It seems like you directly wrote `is_optional` into the single header, so it's missing from the changes.

Smertig:
I didn't change `meta::is_optional`, just propose an idea how my solution can be improved.

ThePhD:
Hey,

thanks for this pull request. I went about it a little differently to gain (a tiny bit) of compiler speed over using struct specialization, but other than that the fixes were mostly identical. I was also super annoyed by `CXX17` feature gates in sol2 since sol2 is already a C++17 library, so while I was at it I blew that up too.

Thanks for the suggestions and help, and sorry it took me so long to get to this one!

PR #978, opened 2020-06-10
correct dot product in example

pbrady:
Not a super important change but some of us are very passionate about dot products.


ThePhD:
kek, nice.

Thanks!

pbrady:
Thanks for merging!

PR #982, opened 2020-06-18
Reduce binary bloat in demangle implementation

btdavis:
ctti_get_type_name is currently ~1kb per instantiation on an optimized build, which adds up quickly with a large number of bindings. This moves the bulk of the implementation into a non-template function to reduce the amount of duplicated code.

Built and tested with MSVC only. I gave my best guess for ifdef'd gcc/clang code but I'm not set up to build with that toolchain at the moment so will need someone to review.

### DIFF
--- a/include/sol/demangle.hpp
+++ b/include/sol/demangle.hpp
@@ -1,4 +1,4 @@
-// sol3 
+// sol3
 
 // The MIT License (MIT)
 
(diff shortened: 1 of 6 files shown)


btdavis:
Added similar improvements for static_trampoline and check_metatable. They were chosen because they had particularly high use in MWSE. All together these changes save about 7MB vs the current MWSE sol3 branch.

Also I found that Sizer (http://aras-p.info/projSizer.html) is the most useful tool for finding these optimizations since it has a "Aggregated templates by size" section in its report.

ThePhD:
This is neato; thanks for this.

I have a lot of changing to do for the sizes and everything.

Don't mind the failed builds: I need to COMPLETELY REDO the Contiguous Integration now that we're officially C++17 and above only.

PR #1003, opened 2020-08-02
no_constructpr, variable name is_const -> ret_is_const

Ubpa:
`is_const` will get some strange messages in VS 2019
maybe it is conflicted with `std::is_const`


ThePhD:
Thanks for this; I have to go through and do a thorough variable rename. I just got done renaming all the macros and using the new tech there, so this should help too!

Appreciate it a lot. :D

PR #1056, opened 2020-11-01
Add CMake option to disable installation

violinyanev:
Rationale: since sol2 is a header-only library, some projects might
use it as a private dependency instead of exposing it through their API.
Thus, it is useful to have an option to not install sol2, in order to
reduce the installation size as well as to avoid accidentally exposing
internal dependencies in the top-level project.

The project I work on is a demonstration of how the feature might be
used:
https://github.com/GENIVI/ramses-logic/blob/f4a20ec39ea2a29e9d26a82564d15c4d6e2405a7/external/CMakeLists.txt#L84
(mind the different name of the CMake variable, I renamed it to fit the rest
of the sol options convention)

We have an install-check script script which verifies that only expected
headers, libs, targets and docs are exported (the one which are interesting
from users PoV). I wouldn't mind contributing the script as github workflows
if you want to use it as well - please provide feedback if that's desired.
Here is a link to our script for preview:
https://github.com/GENIVI/ramses-logic/blob/master/ci/scripts/installation-check/check-installation.py

While modifying the CMake code, I noticed a 4-space instead of TAB in one
place and fixed it to be consistent with the rest of the CMake code.

Tested locally with all examples, tests etc. and it seems to work - no INSTALL target appears on VS2019. Couldn't test on
Linux because I don't have a setup yet, and didn't test with DOCs enabled (no sphinx, and python venv is a bitch to setup
on windows...).


ThePhD:
Hey this looks great! You don't have to worry about fixing the appveyor.yml and stuff; I'm in the process of getting rid of it in favor of GitHub Actions. The rest looks okay, and thanks for catching that formatting inefficiency!

`SOL2_ENABLE_INSTALL` defaulting to `ON` is also the right thing to do here. I wonder if there's a more elegant way to mark a dep as "private" and therefore no need to include its e.g. documentation and headers and such...

ThePhD:
But that's outside the scope of what you'd need to do.

Cheers!

violinyanev:
Hi @ThePhD ,
there is a way to mark dependencies to "not install" with CMake generator commands - but it's too complicated and honestly not worth it, apart from it requiring everyone to use it... Not realistic imho :)

Issue #1112, opened 2021-01-21
new_usertype Requires Constructor To Exist Even If sol::no_constructor Is Used


atom0s:
**sol version:** f275ad2fc468365e691e05a944e92310e548dbff
**Lua version:** MoonJIT https://github.com/AshitaXI/moonjit/commit/8827354e537a03e404134a54caa1c8649e46172a

I am working through moving a big project with lots of bindings from Luabind to sol3. I ran into a small snag that I'm not sure is potentially a bug or just a result to how things are done in sol. 

The bindings are within a plugin for a project. The main project owns the objects I am binding to as interfaces / structures. The plugin does not own the types or expose any of the actual method bodies inside of it. However, some of the types do have method definitions to outline calls. Binding these interfaces all work fine, but structs that have an empty constructor do not.

In Luabind, I am able to avoid any issues with this and the constructor is not needed, used or requested at compile time by the bindings. In sol though, this is not the case. Just creating a usertype, even while using `sol::no_constructor` still causes the types constructor to exist in the local codebase.

```cpp
// Example via Luabind, no constructor required and compiles fine..
void setup_binding(lua_State* L)
{
    using namespace luabind;

    module(L)[
        class_<ImGuiIO>(u8"ImGuiIO")

    // bind properties and methods here..
    ];
}

// Example via sol, causes an error complaining ImGuiIO::ImGuiIO does not exist..
void setup_binding(void)
{
    m_LuaState.new_usertype<ImGuiIO>(u8"ImGuiIO"); // errors
    m_LuaState.new_usertype<ImGuiIO>(u8"ImGuiIO", sol::no_constructor); // errors

    // bind properties and methods here..
}
```

` error LNK2001: unresolved external symbol "public: __thiscall ImGuiIO::ImGuiIO(void)" (??0ImGuiIO@@QAE@XZ)`

ThePhD:
Linker errors are not compiler errors, which leads me to believe `ImGuiIO` is still default-constructible (`static_assert(std::is_default_constructible_v<ImGuiIO>)` passes)?

atom0s:
Oops meant to say link time, sorry about that. Yes, that `static_assert` passes.

ThePhD:
This code should now work as expected with `sol::no_constructor` SPECIFICALLY passed in the original `new_usertype` call. Anywhere after and the Linker Error will still come uppercut you.

atom0s:
Looks like current develop is broken from compiling even a basic example.

[...]

PR #1113, opened 2021-01-21
Reorder some usertype_storage usage to fix a compile error.

atom0s:
Fixes compile errors with the usertype_storage header by reordering some of the usages. 

Fixes the issue seen here: https://github.com/ThePhD/sol2/issues/1112#issuecomment-764943657


ThePhD:
I fixed this while fixing a bunch of other stuff. Sorry!

PR #1139, opened 2021-02-21
Add sol2 include directories as system includes

FlexW:
If someone enables warnings as errors on their target and links
against sol2, the errors/warnings in sol2 prevents that the target
gets compiled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ set_target_properties(sol2
 	PROPERTIES
 	EXPORT_NAME sol2::sol2)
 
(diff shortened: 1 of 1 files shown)


ThePhD:
On one hand, I don't want to be a system include because this should come /W4 and -Wall clean.

On the other hand, reality is a thing and I need to stop living in fantasy land.

I didn't accept this, but I did turn an on-by-default "SOL2_SYSTEM_INCLUDE" CMake variable option for this. So, effectively, you still don't have to do anything.

FlexW:
Thank you:)

PR #1193, opened 2021-05-10
Fix aligned memory allocation

### LINKED ISSUE #1192: Aligned allocation of userdata block (data section) for `Test` failed
Hello!

I've found a non-obvious bug in alignment and memory size calculations. Let's take a simple type (assuming x86 build mode):

```cpp
struct Test {
	std::uint64_t dummy;
};
static_assert(sizeof(Test) == 8);
static_assert(alignof(Test) == 8);
static_assert(sizeof(Test*) == 4);
static_assert(alignof(Test*) == 4);
```

The following piece of code leads to panic (_sometimes_):
```cpp
sol::state lua;
lua["test"] = Test {};
// [sol2] An error occurred and panic has been invoked: aligned allocation of userdata block (data section) for 'Test' failed
```

I've made some research and found the following

-----

The **first** problem is that `misaligned_size` can be less than `aligned_size`:
https://github.com/ThePhD/sol2/blob/f92c28b1b56e85f3b37c84192770d6e4ca49a4a7/include/sol/stack_core.hpp#L344-L345
where `align_space_for` calculates memory size required to place correctly aligned types in memory block starting at specific address.
 - `align_space_for<Test*, Test>((void*)0x0) == 16` (or even 12, see below)
 - `align_space_for<Test*, Test>((void*)0x1) == 15` 

Technically, `0x0` is the best alignment for `Test*` (first type in pack). However, the second type (`Test`) would be placed at minimal address equal to `0x4` or `0x05` but requires `0x8` alignment so there would be a 3-4 byte gap. This means that `align_space_for<...>((void*)0x0)` doesn't calculate required space for best™ allocator and `align_space_for<..>((void*)0x1)` - for the worst™ one. More specifically, the worst case for `Test` is reached when allocator returns pointer with representation `0xXXXXX5`: `align_space_for<Test*, Test>((void*)0x5) == 19`. 

-----

The **second** problem is the attempt to avoid UB, leading to randomization of the result.
https://github.com/ThePhD/sol2/blob/f92c28b1b56e85f3b37c84192770d6e4ca49a4a7/include/sol/stack_core.hpp#L109-L117

First of all, there is an UB for `(void*)0x1` _anyway_. Just because `(void*)0x1` is invalid address and is not `nullptr` so `alignment_shim` won't help.
Secondly, `align_space_for<Test*, Test>(nullptr)` may return `16` or `12` or something else depending on address of a local variable `alignment_shim`:
 - `reinterpret_cast<std::uintptr_t>(&alignment_shim[0]) % 16 == 0` => result is `16`
 - `reinterpret_cast<std::uintptr_t>(&alignment_shim[0]) % 16 == 8` => result is `12`
 - and so on..

In my case `initial_size` is `12`, `misaligned_size` is `15` but allocator returns pointer with alignment equal to `16` requiring at least `16` bytes to place `Test*` and `Test` types.

-----

Tested sol2 version: trunk (f92c28b1b56e85f3b37c84192770d6e4ca49a4a7)

P.S. I'm already working on PR. Hope, you'll accept it.

Smertig:
Closes #1192 

- Reworked `sol::detail::aligned_space_for<Ts...>`
  - If no type is over-aligned (`alignof(Ts) <= alignof(std::max_align_t) for all Ts...`), calculate maximum possible required size assuming that allocator can return absolutely any address
  - Otherwise assume that every single type in pack has initial address `0x1` (the worst alignment)
  - Make it `constexpr`
- Always do only one allocation with required memory size. "aligned allocation failed" error should never happen anymore
- Remove unused arguments/variables (it was probably a legacy code)

### DIFF
--- a/include/sol/stack_core.hpp
+++ b/include/sol/stack_core.hpp
@@ -78,7 +78,7 @@ namespace sol {
 #endif
 		}
 
-		inline std::uintptr_t align(std::size_t alignment, std::uintptr_t ptr, std::size_t& space) {
+		constexpr std::uintptr_t align(std::size_t alignment, std::uintptr_t ptr, std::size_t& space) {
 			// this handles arbitrary alignments...
 			// make this into a power-of-2-only?
 			// actually can't: this is a C++14-compatible framework,
@@ -94,13 +94,13 @@ namespace sol {
 			return reinterpret_cast<void*>(align(alignment, reinterpret_cast<std::uintptr_t>(ptr), space));
 		}
 
-		inline std::uintptr_t align_one(std::size_t alignment, std::size_t size, std::uintptr_t target_alignment) {
+		constexpr std::uintptr_t align_one(std::size_t alignment, std::size_t size, std::uintptr_t target_alignment) {
 			std::size_t space = (std::numeric_limits<std::size_t>::max)();
 			return align(alignment, target_alignment, space) + size;
 		}
 
 		template <typename... Args>
-		std::size_t aligned_space_for(std::uintptr_t alignment) {
+		constexpr std::size_t aligned_space_for(std::uintptr_t alignment) {
 			std::uintptr_t start = alignment;
 			(void)detail::swallow { int {}, (alignment = align_one(std::alignment_of_v<Args>, sizeof(Args), alignment), int {})... };
 			return static_cast<std::size_t>(alignment - start);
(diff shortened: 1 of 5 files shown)


ThePhD:
This actually does look good to me, albeit the tests will always fail for one or two platforms because we're running into a hard regression thanks to `lua_error` thrashing the stack in some cases of optimized code. Can't escape it, so the CI will always go red.

I'm gonna run this locally and try to fix things up, so forgive me if merging this takes a little bit!!!

ThePhD:
Lllllooooooooooooooooooooooks good to me and makes things much faster / more consistent when alignment gets weird. Thanks so much for this!

I was originally going to promote these functions out of the `detail` namespace since they're used but I'll do that later, since that's technically a breaking change.

ThePhD:
Ssssskskskskdskdjakdasjk I closed it and almost panicked because the merge button disappeared

I AM A PROFESSIONAL I PROMISE!!

PR #1194, opened 2021-05-13
Added FFI enabled check

RinatNamazov:
Fixes "undefined reference" error if LuaJIT is built with FFI disabled.

### DIFF
--- a/include/sol/compatibility/lua_version.hpp
+++ b/include/sol/compatibility/lua_version.hpp
@@ -70,6 +70,14 @@
 	#define SOL_LUAJIT_VERSION_I_ 0
 #endif
 
+#if defined(SOL_LUAJIT_FFI_DISABLED)
+	#define SOL_LUAJIT_FFI_DISABLED_I_ SOL_ON
+#elif defined(LUAJIT_DISABLE_FFI)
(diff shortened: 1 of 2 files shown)


ThePhD:
Hey, thanks for the patch!

Typically, what we do is we create a special macro post-fixed with `_I_` that sits in the `version.hpp` file. That special macro is "internal", and is triggered by build-based defines. Because this one comes from LuaJIT, you might need to add it to the `compatibility.hpp` file where a few other configuration defines are put in.

The reason we do this is to allow the user to manually specify it, if they want to. If the user doesn't, then it falls back to default logic. After that, we use the value produced there in the checks in the real code, so it'd be

```cpp
#if SOL_IS_ON(SOL_USE_LUAJIT_I_) && SOL_IS_OFF(SOL_LUAJIT_FFI_DISABLED_I_)
```

ThePhD:
Beautiful, thank you so much!

vdweller84:
Single header seems to be missing this. I patched it using

```
		#ifndef SOL_LUAJIT_FFI_DISABLED
						luaL_requiref(L, "ffi", luaopen_ffi, 1);
						lua_pop(L, 1);
		#endif
```

Let me know if anything will blow up doing this.

ThePhD:
That honestly looks okay to me.

PR #1196, opened 2021-05-18
Fix sol2 cmake include failing as a dependency

edunad:
When using cmake `FetchContent` sol isn't the top level project, it should use `CMAKE_CURRENT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR`


OrfeasZ:
Came across this issue as well when using sol with `add_subdirectory`.

PR #1277, opened 2021-11-12
Fix compile error when registering new user types with operator()

matusfedorko:
Compiling with the latest develop branch, the following would fail to compile (tested in MSVC 2019):

```
#define SOL_ALL_SAFETIES_ON 1
#include <sol/sol.hpp>

struct Ptr
{
	int* operator()() { return nullptr; }
	const int* operator()() const { return nullptr; }
};

int main()
{
	sol::state lua;
	lua.new_usertype<Ptr>("Ptr");
	return 0;
}
```

### DIFF
--- a/include/sol/bind_traits.hpp
+++ b/include/sol/bind_traits.hpp
@@ -31,7 +31,7 @@
 namespace sol { namespace meta {
 	namespace meta_detail {
 		template <typename F>
-		using detect_deducible_signature = decltype(&F::operator(), void());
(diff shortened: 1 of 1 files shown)


ThePhD:
I think this should work, but I need to run the tests for a bunch of platforms to make sure.

I also need to get GitHub Actions running and figure out how to rip these appveyor/travis-ci checks out of the PRs already...

ThePhD:
We're about to find out how broken this is! Thanks for submitting.

PR #1295, opened 2022-01-03
Fixed: `std::cout` not found on msvc

Klaim:
I got that error (`std::cout` doesnt exist) while trying to build `develop` branch with MSVC (2022) after having added the regression tests to the `build2` package. Could be fixed some other ways.


ThePhD (approved):

ThePhD:
Thank you!

PR #1352, opened 2022-05-09
CMake: Build LuaJIT: use copy_if_different

DerSauron:
Use copy_if_different to move luajit products in the post build step to
avoid relinking of projects depending on sol2/luagit.
This is mostly relevant when building sol2 and luajit as subprojects.


ThePhD:
Good idea: thanks for submitting this!

PR #1364, opened 2022-06-01
Fixed double deletion of "const reference" usertypes returned from functions

erikogenvik:
Fixes for the issue where const references are incorrectly deleted by Lua GC, as described in a separate Issue (https://github.com/ThePhD/sol2/issues/1365).

This fixes the issue, and all tests passes, but it's not entirely clear to me what the impact is, and why the is_const check in push_reference was added to start with.

Note that I had to upgrade Clutch2 to handle an issue with building on glibc 2.34 (I'm on OpenSuSE Tumbleweed which uses 2.35).
I also had to handle compile warnings on gcc 12.1 which made compilation fail (which hopefully won't blow up on other compilers).
And since I'm on Linux I added a fix for generating Windows file ending for the single header.

### DIFF
--- a/cmake/Modules/FindCatch.cmake
+++ b/cmake/Modules/FindCatch.cmake
@@ -28,7 +28,7 @@ include(Common/Core)
 if (Catch_FIND_VERSION)
 	set(catch_version ${Catch_FIND_VERSION})
 else()
-	set(catch_version 2.11.1)
+	set(catch_version 2.13.8)
 endif()
 
 set(catch_lib catch_lib_${catch_version})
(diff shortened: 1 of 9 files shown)


ThePhD:
I'm not sure if this is correct, because the point of `const` references is to copy them in (unless someone does `std::reference_wrapper<const T>`, in which case we have to make a reference no matter what).

But it's been a bit and I've intended to rewrite this sytem anyways to allow for both const and non-const references to exist within Lua and have their metatables properly restricted.

ThePhD:
Going to close this one, since it's in semi-conflict with what we're trying to fix upstream.

PR #1395, opened 2022-08-26
Add to_string and is_integer utilities

gotschmarcel-ni:
I've been using sol for quite a while now and these two utilities can sometimes be handy to have in C++. I'm not a Lua C-API expert, so it might not be the best possible solution or contain bugs. I'm happy to address suggestions if such an addition is considered useful for others too.

### DIFF
--- a/tests/run_time/CMakeLists.txt
+++ b/tests/run_time/CMakeLists.txt
@@ -1,18 +1,18 @@
 # # # # sol2
 # The MIT License (MIT)
-# 
+#
 # Copyright (c) 2013-2022 Rapptz, ThePhD, and contributors
-# 
+#
 # Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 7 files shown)


tests/run_time/CMakeLists.txt
```diff
@@ -22,12 +22,12 @@
...
 
-file(GLOB sources
-	LIST_DIRECTORIES FALSE
-	CONFIGURE_DEPENDS
-	source/*.cpp)
-
-if (WAY_TOO_BUSTED_RIGHT_NOW)
+# Many are broken, so we list the working ones manually for now.
+set(sources
+	source/is_integer.cpp
```
gotschmarcel-ni, line 27: I didn't test which other files might also still work. I can of course add other files too, if you tell which ones should work.

ThePhD: I think this should not be touched and I should focus on fixing the tests in totality..

PR #1428, opened 2022-12-13
Allow compiling with the CUDA 11.4 compiler

ibaned:
Hello,

This pull request allows sol2 to be compiled by version 11.4 of the CUDA C++ compiler `nvcc`. It essentially consists of workarounds for compiler bugs, and I certainly understand the annoyance of changing one's code due to bad compilers, but if this is merged then the mainline version of sol2 will be able to support our High Performance Computing codes with no modifications.

Thank you for any time you spend reviewing this and considering it for merging, please let me know if there is anything I can do to help.

I don't think it should change any behavior, although the construction of a pointer by forwarded arguments change might be worth a look by someone with deeper C++ knowledge.

### DIFF
--- a/include/sol/stack_core.hpp
+++ b/include/sol/stack_core.hpp
@@ -1190,7 +1190,7 @@ namespace sol {
 		}
 
 		template <typename T>
-		decltype(auto) get_usertype(lua_State* L, int index = -lua_size_v<meta::unqualified_t<T>>) {
+		decltype(auto) get_usertype(lua_State* L, int index = -lua_size_v<meta::unqualified_t<T> >) {
 			record tracking {};
 			return get_usertype<T>(L, index, tracking);
 		}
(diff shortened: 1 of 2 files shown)


include/sol/stack_core.hpp
```diff
@@ -1190,7 +1190,7 @@ namespace sol {
 		}
 
 		template <typename T>
-		decltype(auto) get_usertype(lua_State* L, int index = -lua_size_v<meta::unqualified_t<T>>) {
+		decltype(auto) get_usertype(lua_State* L, int index = -lua_size_v<meta::unqualified_t<T> >) {
```
jarekpelczar, line 1193: Maybe these changes should require some clang-format suppressions, otherwise it might be wiped in the next instance ?

ibaned:
For the time being we've decided to go a different route that doesn't use sol2, so I'll close this to not take up developer time.

PR #1446, opened 2023-01-25
version: detects clang as MSVC and GCC

gotschmarcel-ni:
sol uses `__GNUC__` to detect GCC and `_MSC_VER` to detect MSVC. The problem is that clang defines both of them.

### DIFF
--- a/include/sol/version.hpp
+++ b/include/sol/version.hpp
@@ -95,7 +95,7 @@
 	#else
 		#define SOL_COMPILER_VCXX_I_ SOL_OFF
 	#endif
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) && !defined(__clang__)
 	#define SOL_COMPILER_VCXX_I_ SOL_DEFAULT_ON
 #else
 	#define SOL_COMPILER_VCXX_I_ SOL_DEFAULT_OFF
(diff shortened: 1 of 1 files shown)


ThePhD:
This is fine. We want this to happen: you can know if it's a mixed-clang by checking `SOL_IS_ON(SOL_COMPILER_VCXX)` with `SOL_IS_ON(SOL_COMPILER_CLANG)`, same with GCC.

Albeit, we should *probably* improve the GCC detection to be a lot more specific to GCC so it doesn't fire on Clang.

ThePhD:
Ah, no. It's fine as is. If there's any place where we should specifically be checking for Clang first before GCC, that might help push me in one direction or another.

gotschmarcel-ni:
I don't remember the exact error right now, but for us this was causing compile errors.

gotschmarcel-ni:
Just compiled again w/o our local patch and we get the following error. Note besides, we use `-Werror`

```
sol-v3.3.0/include/sol/stack_check_get_qualified.hpp:35:32: error: unknown warning group '-Wmaybe-uninitialized', ignored [-Werror,-Wunknown-warning-option]
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```

PR #1448, opened 2023-01-31
Add note for LuaJIT 5.2 style containers in the documentation

tatjam:
If you use the `LUAJIT_ENABLE_LUA52COMPAT` compilation flag with LuaJIT, C++ containers can be used just fine with `ipairs` or `pairs` as these properly check the metatable keys `__ipairs` and `__pairs` used by sol to implement this behavior.

I've not tested this behavior in-depth, but it should work fine with all containers.


ThePhD:
This is great to know; thank you!

PR #1454, opened 2023-02-10
Fix cpp_object usertype error

likun123687:
Fix usertype_call_from_c++.cpp cpp_object usertype error


ThePhD:
Wait. The rest of the code still uses e.g. `lua["test"]`, though. This would cause things to break...?

I'll adjust it locally and re-up this though, I guess.

ThePhD:
This is fixed properly now, thanks.

PR #1472, opened 2023-04-05
Fix build when LUA_COMPAT_BITLIB or LUA_COMPAT_5_2 are defined

### LINKED ISSUE #1461: Not possible to build sol2 v3.3.0 with Lua 5.3 with bitlib compatibility enabled
The issue is an apparent misinterpretation of how Lua compatibility macros are defined.

Consider this fragment from Lua’s src/luaconf.h (starts at line 303 in Lua 5.3.4):
```c
#if defined(LUA_COMPAT_5_2)	/* { */

/*
@@ LUA_COMPAT_MATHLIB controls the presence of several deprecated
** functions in the mathematical library.
*/
#define LUA_COMPAT_MATHLIB

/*
@@ LUA_COMPAT_BITLIB controls the presence of library 'bit32'.
*/
#define LUA_COMPAT_BITLIB

/*
@@ LUA_COMPAT_IPAIRS controls the effectiveness of the __ipairs metamethod.
*/
#define LUA_COMPAT_IPAIRS

/*
@@ LUA_COMPAT_APIINTCASTS controls the presence of macros for
** manipulating other integer types (lua_pushunsigned, lua_tounsigned,
** luaL_checkint, luaL_checklong, etc.)
*/
#define LUA_COMPAT_APIINTCASTS

#endif				/* } */
```

Note that if Lua 5.2 compatibility is enabled, `LUA_COMPAT_BITLIB` will be defined as an empty macro.

Now consider this code in sol2’s include/sol/compatibility/lua_version.hpp (beginning at line 186 in sol2 v3.3.0):
https://github.com/ThePhD/sol2/blob/v3.3.0/include/sol/compatibility/lua_version.hpp#L186
```c++
#if defined (SOL_LUA_BIT32_LIB)
	#if SOL_LUA_BIT32_LIB != 0
		#define SOL_LUA_BIT32_LIB_I_ SOL_ON
	#else
		#define SOL_LUA_BIT32_LIB_I_ SOL_OFF
	#endif
#else
	// Lua 5.2 only (deprecated in 5.3 (503)) (Can be turned on with Compat flags)
	// Lua 5.2, or other versions of Lua with the compat flag, or Lua that is not 5.2 with the specific define (5.4.1 either removed it entirely or broke it)
	#if (SOL_LUA_VERSION_I_ == 502) || (defined(LUA_COMPAT_BITLIB) && (LUA_COMPAT_BITLIB != 0)) || (SOL_LUA_VERSION_I_ < 504 && (defined(LUA_COMPAT_5_2) && (LUA_COMPAT_5_2 != 0)))
		#define SOL_LUA_BIT32_LIB_I_ SOL_ON
	#else
		#define SOL_LUA_BIT32_LIB_I_ SOL_DEFAULT_OFF
	#endif
#endif
```

On line 195, this fragment causes a compilation error with clang and GCC: `(defined(LUA_COMPAT_BITLIB) && (LUA_COMPAT_BITLIB != 0))`

Because `LUA_COMPAT_BITLIB` is defined as an empty macro, it’s substituted as `(1 && ( != 0))` so the compilers produce an error because there’s no left-hand argument to the not-equal operator (`!=`).

The `LUA_COMPAT_*` macros should only be tested for being defined or undefined.  The Lua headers define them as empty macros when enabled, so attempting to compare them to 1/0 causes compilation errors.

I can open a PR to change line 195 to this if it would be considered a good solution:
```c++
	#if (SOL_LUA_VERSION_I_ == 502) || defined(LUA_COMPAT_BITLIB) || (SOL_LUA_VERSION_I_ < 504 && defined(LUA_COMPAT_5_2))
```

lmoureaux:
Hi and thank you for sol2!

Lua may define the macros but leaves them empty. The code assumed they would contain a boolean value.

This is required to use sol2 with system Lua on Fedora 37 (see https://github.com/longturn/freeciv21/issues/1895)

Closes #1461.

### DIFF
--- a/include/sol/compatibility/lua_version.hpp
+++ b/include/sol/compatibility/lua_version.hpp
@@ -192,7 +192,7 @@
 #else
 	// Lua 5.2 only (deprecated in 5.3 (503)) (Can be turned on with Compat flags)
 	// Lua 5.2, or other versions of Lua with the compat flag, or Lua that is not 5.2 with the specific define (5.4.1 either removed it entirely or broke it)
-	#if (SOL_LUA_VERSION_I_ == 502) || (defined(LUA_COMPAT_BITLIB) && (LUA_COMPAT_BITLIB != 0)) || (SOL_LUA_VERSION_I_ < 504 && (defined(LUA_COMPAT_5_2) && (LUA_COMPAT_5_2 != 0)))
+	#if (SOL_LUA_VERSION_I_ == 502) || defined(LUA_COMPAT_BITLIB) || (SOL_LUA_VERSION_I_ < 504 && defined(LUA_COMPAT_5_2))
 		#define SOL_LUA_BIT32_LIB_I_ SOL_ON
 	#else
 		#define SOL_LUA_BIT32_LIB_I_ SOL_DEFAULT_OFF


llalonde1:
Is there an update on this?

ThePhD:
I don't think this is a valid PR. I absolutely want to check if the value is defined to 0, and not act on it if it's defined to be 0. 0 means "off", not-0 means "on".

cuavas:
You’re misunderstanding how Lua uses feature macros.  As detailed in #1461, Lua does not define these macros to 0 or 1 to disable or enable the feature.  The feature is enabled whenever the macro is defined at all, and src/luaconf.h defines `LUA_COMPAT_BITLIB` as an empty macro.

ThePhD:
I'm in a bit of a bind, then, because some folk are using custom confs to explicitly shut it off by redefining it to be 0.... well.

Maybe this just means those folks will have to switch to agressive `#undef`. Hell if I know what's the right way anymore.

ThePhD:
Either way, this is back to a way that'll work for everyone in this PR by way of this: https://github.com/ThePhD/sol2/commit/839f2a1dceb0c274d54ba755b2b508df3b7fdca5

PR #1573, opened 2024-01-13
Print aux_message if there is no message

loveridge:
The aux_message would not be printed if the message was empty because the fstring `%s` is filled with the empty message argument.

Before: `stack index 2, expected string, received no value:`

After: `stack index 2, expected string, received no value: (bad argument into 'sol::basic_table_core<false, sol::basic_reference<false> >(Rml::Context&, const std::__cxx11::basic_string<char>&, sol::basic_object<sol::basic_reference<false> >, sol::this_state)') `


ThePhD:
Good catch; thanks for submitting!

PR #1590, opened 2024-03-22
Add limits header explicitly to be compatible with gcc11

JKraso:
Based on https://gcc.gnu.org/gcc-11/porting_to.html `Header dependency changes`, new versions `>gcc10` have removed some implicit includes (again) which causes codebases relying on those to fail.

This little line will fix compatibility with gcc11

### DIFF
--- a/include/sol/stack_core.hpp
+++ b/include/sol/stack_core.hpp
@@ -45,6 +45,7 @@
 #include <sstream>
 #include <optional>
 #include <type_traits>
+#include <limits>
 
 namespace sol {
 	namespace detail {


Rochet2:
The header is already added above on line 43. It was added in this commit https://github.com/ThePhD/sol2/commit/e5e6466e09b632677d24a8f204d6a0ea0a8862b1

How did you get the error? Were you using latest sol? (v3.3.0)

JKraso:
Wow ok. I'm using CPM and even though it told me it's version 3.3.0 i guess it was still the old one. Sorry about that.

PR #1606, opened 2024-05-25
fix emplace

paulocoutinhox:


paulocoutinhox:
Hi, can you merge? Thanks.

pinguin999:
I don't understand the problem, but this code part blocks the build with the latest version of emscripten.

hsdk123:
> I don't understand the problem, but this code part blocks the build with the latest version of emscripten.

Just to give more context, the error https://github.com/ThePhD/sol2/issues/1611 here is shown with the latest version of emscripten.

This PR seems to fix it.

cuavas:
I don’t think this implementation of `emplace` makes sense for `optional<T&>` as written.

Consider:
```c++
	template <class T>
	class optional<T&> {
...
		template <class... Args>
		T& emplace(Args&&... args) noexcept {
			static_assert(std::is_constructible<T, Args&&...>::value, "T must be constructible with Args");

			*this = nullopt;
			new (static_cast<void*>(this)) optional(std::in_place, std::forward<Args>(args)...);
			return **this;
		}
...
	};
```

Now look at the `static_assert` – it’s testing that `T` can be constructed using the argument types passed to `emplace`.  However, the `optional<T&>` doesn’t hold `T`, it logically holds `T&` (although it’s implemented using `T*` to allow assignment).  The only thing that `T&` can be constructed with is `T&`.  It makes no sense to use variadic arguments.

When called with `T&`, the `static_assert` only succeeds by chance when `T` is copy constructible.  It will fail if it isn’t, and it will succeed for things it shouldn’t.

Since you know `T*` is trivially destructible and copyable, you could implement `emplace` as:
```c++
		T& emplace(T& arg) noexcept {
			m_value = &arg;
			return **this;
		}
```

MrG9090 (approved):

PR #1632, opened 2024-10-02
usertype_container.hpp: fix bugs introduced by previous refactoring

### LINKED ISSUE #1627: Automatic container wrapper is not working in devel branch
I tried develop branch because there is a fix for passing nil from lua to C function expecting shared_ptr (or sol::object) using luajit as VM.
It is fixed in commit ff3f254f7b47cfc8cf2976368d86f4cd726915ec 

However it looks like commit 64e3823bd200913d3e92123cbb2e76d9ec377d3e breaks the default container  wrapper.

Simple example:
```c++
#include <map>
#include <sol/sol.hpp>


int main() {
    std::map<std::string, bool> m;

    sol::state state;
    state["m"] = m;
}
```
System: linux, clang, luajit

Output:
[output.log](https://github.com/user-attachments/files/16994048/output.log)

deadlocklogic:
Closes #1627.
Notice that the unit tests **passed** before this fix **unexpectedly**, see the related open issue.

### DIFF
--- a/include/sol/usertype_container.hpp
+++ b/include/sol/usertype_container.hpp
@@ -1188,8 +1188,8 @@ namespace sol {
 			template <bool ip>
 			static int next_associative(std::true_type, lua_State* L_) {
 				iter& i = stack::unqualified_get<user<iter>>(L_, 1);
-				auto& it = i.it;
-				auto& end = i.end;
+				auto& it = i.it();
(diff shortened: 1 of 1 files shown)


ThePhD:
Sorry for the delay.

This is part of latest v3.5.0 -- thanks for catching this issue! This closes #1627.

PR #1654, opened 2024-11-29
Fix issue 1643 shared_ptr<const TYPE>& as parameter not working

### LINKED ISSUE #1643: Calling functions which receive `shared_ptr<const TYPE>&` or `const shared_ptr<const TYPE>&` doesn't work
Calling functions which receive `shared_ptr<const TYPE>&` or `const shared_ptr<const TYPE>&` doesn't work:
```
#include <iostream>
using namespace std;

#include "lua.hpp"

#define SOL_ALL_SAFETIES_ON 1
#include "sol3.hpp"

int main() {
    sol::state lua;
    lua.set_function("intBuilder", []() -> shared_ptr<const int> { return make_shared<int>(10); });
    lua.set_function("intReader", [](shared_ptr<const int>& val) { cout << "intReader" << endl;} );
    lua.script(R"(
               print('hello from lua')
               local sharedInt = intBuilder()
               intReader(sharedInt)
               )");
    return 0;
}
```
The output I receive:
```
hello from lua
[sol2] An error occurred and has been passed to an error handler: sol: runtime error: stack index 1, expected userdata, received sol.sol::d::u<int>: unrecognized userdata (not pushed by sol?) (bad argument into 'void(std::shared_ptr<const int>&)')
stack traceback:
        [C]: in function 'base.intReader'
        [string "..."]:4: in main chunk
terminate called after throwing an instance of 'sol::error'
  what():  sol: runtime error: stack index 1, expected userdata, received sol.sol::d::u<int>: unrecognized userdata (not pushed by sol?) (bad argument into 'void(std::shared_ptr<const int>&)')
stack traceback:
        [C]: in function 'base.intReader'
        [string "..."]:4: in main chunk
Aborted (core dumped)
```
I use gcc 11.4.0 and sol2 v3.3.0 (revision eba86625), C++17.

For [v2.20.6] it's even worse as not even passing by value a `shared_ptr<const TYPE>` works.

I need to map some C++ interfaces to Lua and cannot change those interfaces. Many functions in those interfaces receive `const shared_ptr<const TYPE>&`.

For sol2 v3.3 I can work around this with a copy in a wrapped function and for 2.20.6, additionally I am wrapping the shared pointer into another struct and unwrapping that through the interfaces I need.

Can we get this fixed in the library or is there another 'simpler in some sense' way to work around this problem?
Am I missing something here?

Thank you

EugenCazacu:
Fix #1643

### DIFF
--- a/include/sol/stack_check_unqualified.hpp
+++ b/include/sol/stack_check_unqualified.hpp
@@ -218,6 +218,7 @@ namespace sol { namespace stack {
 			}
 			else if constexpr (is_unique_usertype_v<T>) {
 				using element = unique_usertype_element_t<T>;
+				using element_no_cv = meta::unqualified_t<element>;
 				using actual = unique_usertype_actual_t<T>;
 				const type indextype = type_of(L_, index);
 				tracking.use(1);
@@ -229,7 +230,7 @@ namespace sol { namespace stack {
 					return true;
 				}
 				int metatableindex = lua_gettop(L_);
-				if (stack_detail::check_metatable<d::u<element>>(L_, metatableindex)) {
+				if (stack_detail::check_metatable<d::u<element_no_cv>>(L_, metatableindex)) {
 					void* memory = lua_touserdata(L_, index);
 					memory = detail::align_usertype_unique_destructor(memory);
 					detail::unique_destructor& pdx = *static_cast<detail::unique_destructor*>(memory);


ThePhD:
Sorry for the delay.

This is part of latest v3.5.0 -- thanks for the contribution!

PR #1662, opened 2025-01-02
better header file

cxwx:
for most lua version header file is include in "lua/lua.h" not "lua.h"


ThePhD:
Sorry for the delay.

This is part of latest v3.5.0 (but in a different manner) -- thanks for the contribution!

PR #1676, opened 2025-03-11
Change end() to sen() in usertype_container.hpp

nyllet:
i need to change this to be able to compile. i'm using g++-14 on debian linux


DarthGandalf:
Confirmed to fix https://bugs.gentoo.org/955999 for me, thanks

solbjorn:
I can confirm this fixes the compilation when exporting `absl::btree_map` as a container, thanks!

Issue #1581, opened 2024-02-16
Doesn't build with Clang 18.1.0


checkraisefold:
On the latest major release (candidate) of Clang, sol2 flat out doesn't compile with a fatal error and spits out a bunch of warnings.
https://github.com/jpxs-intl/RosaServer/actions/runs/7929791195/job/21650728959 for the run logs, or here:
```
/home/runner/work/RosaServer/RosaServer/sol2/include/sol/function_types.hpp:110:31: fatal error: address of overloaded function 'call' does not match required type 'int (lua_State *)'
  110 |                                 lua_CFunction freefunc = &function_detail::upvalue_this_member_variable<C, Fx>::template call<is_yielding, no_trampoline>;
      |                                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/RosaServer/RosaServer/sol2/include/sol/function_types.hpp:268:5: note: in instantiation of function template specialization 'sol::function_detail::select_member_variable<false, false, float Vector::*>' requested here
  268 |                                 select_member_variable<is_yielding, no_trampoline>(L, std::forward<Fx>(fx), std::forward<Args>(args)...);
      |                                 ^
/home/runner/work/RosaServer/RosaServer/sol2/include/sol/function_types.hpp:392:22: note: in instantiation of function template specialization 'sol::function_detail::select<false, false, float Vector::*>' requested here
  392 |                                 function_detail::select<false, false>(L, std::forward<Args>(args)...);
      |                                                  ^
/home/runner/work/RosaServer/RosaServer/sol2/include/sol/stack_core.hpp:878:14: note: in instantiation of function template specialization 'sol::stack::unqualified_pusher<float Vector::*>::push<float Vector::*>' requested here
  878 |                                 return p.push(L, std::forward<T>(t), std::forward<Args>(args)...);
      |                                          ^
/home/runner/work/RosaServer/RosaServer/sol2/include/sol/stack_field.hpp:222:7: note: in instantiation of function template specialization 'sol::stack::push<float Vector::*>' requested here
  222 |                                                 push(L, std::forward<Value>(value));
      |                                                 ^
[...]

PR #1722, opened 2025-07-20
Work around #1581

glebm:
See https://github.com/ThePhD/sol2/issues/1581#issuecomment-2103463524

/cc @cuavas

### DIFF
--- a/include/sol/function_types_stateless.hpp
+++ b/include/sol/function_types_stateless.hpp
@@ -320,7 +320,13 @@ namespace sol { namespace function_detail {
 		}
 
 		template <bool is_yielding, bool no_trampoline>
-		static int call(lua_State* L) noexcept(std::is_nothrow_copy_assignable_v<T>) {
+		static int call(lua_State* L)
+#if SOL_IS_ON(SOL_COMPILER_CLANG) && !(__clang_major__ > 19 || (__clang_major__ == 19 && __clang_minor__ >= 1))
+		// apparent regression in clang 18 - llvm/llvm-project#91362
+#else
+			noexcept(std::is_nothrow_copy_assignable_v<T>)
+#endif
+		{
 			int nr;
 			if constexpr (no_trampoline) {
 				nr = real_call(L);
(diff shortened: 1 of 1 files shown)


include/sol/function_types_stateless.hpp
```diff
@@ -320,7 +320,13 @@ namespace sol { namespace function_detail {
 		}
 
 		template <bool is_yielding, bool no_trampoline>
-		static int call(lua_State* L) noexcept(std::is_nothrow_copy_assignable_v<T>) {
+		static int call(lua_State* L)
+#if SOL_IS_ON(SOL_COMPILER_CLANG)
+		// apparent regression in clang 18 - llvm/llvm-project#91362
```
Nerixyz: Clang 19.1+ doesn't have this bug anymore, so this should probably check for Clang 18 only.

glebm: Do you know how to do that? I can amend the PR. I'm guessing there should be a `clang_version` macro or something

glebm: I guess `!(__clang_major__ > 19 || (__clang_major__ == 19 && __clang_minor__ >= 1))`?

Razakhel: This check is too broad, it will [exclude all versions of Clang below 19.1](https://godbolt.org/z/53TGvdobj), not just the `[18.X; 19.1[` range. It [should rather be](https://godbolt.org/z/qz7jbfz4P):
```cpp
// All versions below 18, above 19, or 19.1+
#if !SOL_IS_ON(SOL_COMPILER_CLANG) || __clang_major__ < 18 || __clang_major__ > 19 || (__clang_major__ == 19 && __clang_minor__ >= 1)
			noexcept(std::is_nothrow_copy_assignable_v<T>)
#endif
```

Or [more specific](https://godbolt.org/z/Tx8WfoPP5):
```cpp
// Just not the 18.X and 19.0
#if !SOL_IS_ON(SOL_COMPILER_CLANG) || __clang_major__ != 18 || (__clang_major__ != 19 && __clang_minor__ < 1) // Or __clang_minor__ == 0
			noexcept(std::is_nothrow_copy_assignable_v<T>)
#endif
```

From the comment [here](https://github.com/ThePhD/sol2/issues/1581#issuecomment-2098915311) that gives a minimal example, you can check that [it works just fine for versions below 18](https://godbolt.org/z/7nTs8rczP).

Issue #1721, opened 2025-07-17
Support Lua 5.5


s13n:
Lua 5.5 is in beta, and soon will be released. See https://www.lua.org/work/

Leadwerks:
The global variables feature looks interesting...

cxwx:
#include <sol/sol.hpp>↴    ■ In included file: "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"

PR #1723, opened 2025-07-21
Add lua 5.5 support

halx99:
relative issue: #1721


dg0yt:
Additional changes are needed in `sol/compatibility/compat-5.3.h`.

halx99:
> Additional changes are needed in `sol/compatibility/compat-5.3.h`.

done, thanks

OgreTransporter:
https://github.com/ThePhD/sol2/blob/c1f95a773c6f8f4fde8ca3efe872e7286afe4444/CMakeLists.txt#L242

The switch for 5.5 is missing. Also CMake 4.3 is required.

halx99:
> https://github.com/ThePhD/sol2/blob/c1f95a773c6f8f4fde8ca3efe872e7286afe4444/CMakeLists.txt#L242
> 
> The switch for 5.5 is missing. Also CMake 4.3 is required.

done, thanks.

OgreTransporter:
😄 That doesn't work that way.
```
cmake_minimum_required(VERSION 3.26...4.2)
```
Unfortunately, the Lua versions are hard-coded in CMake (https://cmake.org/cmake/help/git-master/module/FindLua.html). If you restrict CMake to 4.2, it will never be able to find Lua 5.5 because Lua 5.5 is only recognized from CMake 4.3 onwards. Leave minimal CMake at 3.26.
```
cmake_minimum_required(VERSION 3.26)
```
Actually, you would have to check the CMake version in the queries and then output an error message if the CMake version is too old. CMake 3.26 recognizes Lua 5.4. For Lua 5.5, I would add the following check:
```
elseif(NORMALIZED_LUA_VERSION MATCHES "5.5")
    if (CMAKE_VERSION VERSION_LESS "4.3")
        message(FATAL_ERROR "Lua 5.5 is only supported from CMake 4.3 onwards! Current version: ${CMAKE_VERSION}")
    endif()
    set(CREATE_LUALIB_TARGET TRUE)
    find_package(Lua 5.5 EXACT REQUIRED)
```

halx99:
But, the latest cmake stable version is 4.2.x

set min cmake required version to latest also not a good idea

Nerixyz:
> 😄 That doesn't work that way.

[`cmake_minimum_required(VERSION <min>[...<policy_max>])`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) does not set a maximum CMake version. As the name suggests, the `policy_max` only affects the policies. Lua 5.5 support in `FindLua.cmake` is not gated behind any policy.

> Actually, you would have to check the CMake version in the queries and then output an error message if the CMake version is too old. CMake 3.26 recognizes Lua 5.4. For Lua 5.5, I would add the following check:

No. For one, users can bring their own `FindLua` and secondly, package managers also hook into the machinery (so you can use 5.5 with a lower CMake version than 4.3). If that's not the case and CMake <4.3 is used, then the built-in `FindLua` should fail - if it doesn't that's not an issue with sol.

OgreTransporter:
> As the name suggests, the policy_max only affects the policies.

That's right, but there's currently no guarantee that a policy won't be implemented. Therefore, it's better to leave the limitation open. This is changing rapidly in CMake development.

> No. For one, users can bring their own FindLua and secondly, package managers also hook into the machinery (so you can use 5.5 with a lower CMake version than 4.3). If that's not the case and CMake <4.3 is used, then the built-in FindLua should fail - if it doesn't that's not an issue with sol.

Then it would make sense to include a corresponding FindLua version with sol. If you first have to build your own FindLua, this could cause problems for many users.

dg0yt:
> Then it would make sense to include a corresponding FindLua version with sol.

You could do that in a separate PR. 
Find modules do more or less informed guessing. This needs maintainance. For a wider effect, someone with love for CMake and Lua should improve the inflexible and undermaintained CMake's `FindLua.cmake`, so that users can just set the input variables as needed, even for new versions of Lua.

dg0yt:
> That's right, but there's currently no guarantee that a policy won't be implemented. Therefore, it's better to leave the limitation open. This is changing rapidly in CMake development.

This doesn't make sense. The policies associated with 4.2 are final with the release of CMake 4.2.
The task associated with bumping max_policy in projects is to actually verify compatibility with that set of policies. Across all platforms and configurations.

include/sol/compatibility/compat-5.3.h
```diff
@@ -405,11 +405,11 @@ COMPAT53_API void luaL_requiref(lua_State *L, const char *modname,
 
 
 /* other Lua versions */
-#if !defined(LUA_VERSION_NUM) || LUA_VERSION_NUM < 501 || LUA_VERSION_NUM > 504
+#if !defined(LUA_VERSION_NUM) || LUA_VERSION_NUM < 501 || LUA_VERSION_NUM > 505
 
-#  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"
+#  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, 5.4 or 5.5)"
```
Saur2000: Maintain the previous grammar:

```suggestion
#  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, 5.4, or 5.5)"
```

include/sol/compatibility/compat-5.3.h
```diff
@@ -405,11 +405,11 @@ COMPAT53_API void luaL_requiref(lua_State *L, const char *modname,
...
 
 /* other Lua versions */
-#if !defined(LUA_VERSION_NUM) || LUA_VERSION_NUM < 501 || LUA_VERSION_NUM > 504
+#if !defined(LUA_VERSION_NUM) || LUA_VERSION_NUM < 501 || LUA_VERSION_NUM > 505
 
-#  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"
+#  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, 5.4 or 5.5)"
 
-#endif /* other Lua versions except 5.1, 5.2, 5.3, and 5.4 */
+#endif /* other Lua versions except 5.1, 5.2, 5.3, 5.4 and 5.5 */
```
Saur2000:
```suggestion
#endif /* other Lua versions except 5.1, 5.2, 5.3, 5.4, and 5.5 */
```

pinguin999:
Works for me. Thanks!

